PR #15, opened 2019-10-08
feat: Updated datasets.OpenFire

frgfm:
This PR aims at:
- updating the OpenFire dataset's reference to a much larger dataset. (#6)
- improving large objects downloading speed with multi-threading. (#14)
- reducing the number of download issues due to temporary issues (#10)

Ongoing bottlenecks:
- server-side errors cannot be handled and the dataset final size will depend on the user (#10)

PR #16, opened 2019-10-09
test: Using module tempfile to avoid failed test on #15

blenzi:


frgfm:
Thanks for the PR!

Try to update your master branch, and create a new branch from it before submitting a PR (to avoid the 6 commits mentioned above)

Regarding the failed test, I'll comment directly in the review!

test/test_datasets.py
```diff
@@ -43,7 +44,7 @@ def test_downloadurls(self):
 
     def test_openfire(self):
 
-        with Path('/tmp') as root:
+        with tempfile.TemporaryDirectory() as root:
```
frgfm: I suggest you check the following [thread](https://stackoverflow.com/questions/51686130/reading-from-a-temporary-directory-in-python-typeerror-expected-str-bytes-or) to leverage the context manager on `TempDirectory` with `pathlib.Path` as the failing line requires a  ̀Path` object.

Otherwise, the test will keep on failing!

frgfm (changes requested):
Please check the comment to fix the failing test

frgfm:
@blenzi Codecov seems to have an issue unrelated to this according to [CI](https://circleci.com/gh/frgfm/PyroNear/125)

The upload of coverage throws:
```
HTTP 400
Commit sha does not match Circle build.
Codecov upload failed
```

I am checking right now how to fix this!

frgfm:
Apparently, it is a known bug: https://github.com/codecov/codecov-bash/issues/134
For next times, when updating your PR, please commit directly to the pulled branch rather than merging into it in order to avoid this issue.

I'll merge the commit for now!

Issue #3, opened 2019-09-26
[references] Add training script for binary classification


frgfm:
## 🚀 Feature
Add training script for binary classification model

## Motivation

In order to have comparable evaluation runs of training, a training script can serve as a template for comparison and reproducibility.

## Pitch

The training script could be called with optional arguments that would tweak the different parameters according to the user's preferences.

blenzi:
When running ```%run references/classification/torch/train.py --data-path data --epochs 0```

Locally (no GPU):
```
AssertionError: Torch not compiled with CUDA enabled
```
In colab (with GPU):
```
AttributeError                            Traceback (most recent call last)

/content/PyroNear/references/classification/torch/train.py in <module>()
    255     args = parser.parse_args()
    256 
--> 257     main(args)

/content/PyroNear/references/classification/torch/train.py in main(args)
    186 
    187     # Scheduler
--> 188     lr_scheduler = optim.lr_scheduler.OneCycleLR(optimizer, max_lr=args.lr,
    189                                               epochs=args.epochs, steps_per_epoch=len(train_loader),
    190                                               cycle_momentum=(not isinstance(optimizer, optim.Adam)),

AttributeError: module 'torch.optim.lr_scheduler' has no attribute 'OneCycleLR'
```

*Suggestion:* unify the default parameters on all scripts. E.g.: `--data-path`:  `./data` in fastai, `./tmp_data` in torch

frgfm:
@blenzi Thanks for the feedback!
**Locally (no GPU)**
Add the following argument `--device cpu` (default is set to cuda)

**In colab (with GPU)**
I have to update the requirements of pytorch for this script (`OneCycleLR` is available since torch 1.3.0 and I guess colab hasn't updated its default version since the release). Try to force install pytorch 1.3.0 in colab for now ;)

Good catch for `--data-path`, I'm updating this right now!

blenzi:
Ok, locally it seems to work after updating torch and:
```%run references/classification/torch/train.py --data-path data --epochs 1 --workers 0 --device cpu```

To avoid the problem on CPU, a default option `--device auto` could be used to detect the presence of a GPU or not.
E.g. for tensorflow: https://stackoverflow.com/questions/38559755/how-to-get-current-available-gpus-in-tensorflow

Could not find immediately an equivalent for torch but trying `torch.cuda` should raise `AttributeError`. I can open an issue later.

PR #23, opened 2019-10-15
feat: Added reference scripts for binary classification

frgfm:
This PR adds a `reference` folder with training scripts for different tasks. More specifically, it provides two scripts for binary classification: one using only the core dependencies of `pyronear` and the other with `fastai`.

Users should note that the `fastai` one actually changes the model's architectures by complexifying the head. Performances are not comparable as is.

See #3


blenzi (changes requested):
I could not make the script ```fastai_train.py``` work neither locally nor on colab.  Locally, first problem:
```
  File "references/classification/fastai_train.py", line 6, in <module>
    from fastai import vision
ModuleNotFoundError: No module named 'fastai'
```
fastai is not in the requirements, I can understand why. But at least a comment in the README or CONTRIBUTING would be useful.

Then:
```
python references/classification/fastai_train.py

Traceback (most recent call last):
  File "references/classification/fastai_train.py", line 77, in <module>
    main(args)
  File "references/classification/fastai_train.py", line 23, in main
    for sample in OpenFire(root=args.data_path, train=False, download=True).data:
  File "/Users/blenzi/cernbox/Workarea/PyroNear/pyronear/datasets/openfire.py", line 54, in __init__
    raise RuntimeError('Dataset not found.' +
RuntimeError: Dataset not found. You can use download=True to download it
```

Since download is already True I am not sure how to proceed. I tried to manually filling up the images folder with the ones that could not be downloaded and got the same error.

Would it be possible to add the scripts (or the functions called) to CI / unittests ?

frgfm:
@blenzi fully agree with the comment on the readme for the requirements! I forgot to add a specific one for the `references` folder. I'll update this accordingly

Regarding the training script, I'm gonna run some tests but I don't see how to add it efficiently to CI. As you pointed out, the dataset download is already long, but since CI only has CPUs, the training (even for one epoch) would be way too long unfortunately...

blenzi:
On the new version, running locally, no GPU:
```
  File "references/classification/fastai_train.py", line 28, in main
    df = pd.DataFrame.from_dict(dict(name=fnames, label=labels, is_valid=is_valid))
NameError: name 'pd' is not defined
```

If I add ```import pandas as pd```:

```
RuntimeError: DataLoader worker (pid 97793) is killed by signal: Unknown signal: 0.
```

Maybe we could add some tests (download, train, test) with a subset of the full dataset to make sure that the whole chain works ?

frgfm:
@blenzi I just pushed a few commits that should your issues.

However, for CI integration, I guess we can test the scripts with `epochs=0` so we can at least check that the loading part is alright. But I see this as a separate topic _(me introducing a mistake in the script vs. testing this at each commit)_, would you mind opening an issue?

We'll flag this issue for the patch after the incoming release `0.1.0` (most likely for `0.1.1`).

blenzi (approved):

PR #25, opened 2019-10-16
Clarify openfire validation set creation

x0s:
This PR aims at:
- clarifying the name to set validation ratio (between 0 and 1) instead of percentage (between 0 and 100)
- raising exception when this ratio is not within correct range. Before, it was possible to set wrong number (ie: setting 21.5 instead of 0.215 would not raise any warning nor exception)

Any feedback is welcome,
cheers

### DIFF
--- a/pyronear/datasets/openfire.py
+++ b/pyronear/datasets/openfire.py
@@ -48,7 +48,12 @@ class OpenFire(VisionDataset):
         self.train = train  # training set or test set
 
         if download:
-            self.download(threads, valid_ratio)
+            if (not isinstance(valid_ratio, (float, type(None))) or
+                not (0 <= valid_ratio <= 1)):
+                raise ValueError(f"ratio to create Validation set should be a floating number between 0 and 1. Ratio found = {valid_ratio}")
+            else:
+                pass
+                self.download(threads, valid_ratio)
 
         if not self._check_exists(train):
             raise RuntimeError('Dataset not found.' +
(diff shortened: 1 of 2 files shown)


pyronear/datasets/openfire.py
```diff
@@ -42,13 +42,16 @@ class OpenFire(VisionDataset):
...
     def __init__(self, root, train=True, transform=None, target_transform=None,
-                 download=False, threads=16, valid_pct=None):
+                 download=False, threads=16, valid_ratio=None):
         super(OpenFire, self).__init__(root, transform=transform,
                                     target_transform=target_transform)
         self.train = train  # training set or test set
 
         if download:
-            self.download(threads, valid_pct)
+            if isinstance(valid_ratio, (float, int)) and not (0 <= valid_ratio <= 1):
```
frgfm, line 51: Shouldn't it be this line be the following instead ?
```
if not isinstance(valid_ratio, (float, int)) or not (0 <= valid_ratio <= 1):
```
Otherwise a `str` object would be passed as `valid_ratio` to `download`. An alternative would be to put the type checking back in  `download`.

test/test_datasets.py
```diff
@@ -72,13 +72,24 @@ def test_openfire(self):
...
             # Check unicity of sample across all splits
             train_paths = [sample['path'] for sample in train_set.data]
             self.assertTrue(all(sample['path'] not in train_paths for sample in test_set.data))
 
+    def test_openfire_validation(self):
+        """Check if Exception is raised when providing incorrect split ratio for validation"""
+
+        with Path(tempfile.TemporaryDirectory().name) as root:
+
+            incorrect_valid_ratios = [-25, -0.01, 1.01, 15]
```
frgfm, line 87: It might be safer to add non-numerical values to test the behaviour from the previous comment.

frgfm (changes requested):
PR looks great, there is only a type checking that should be put back or modified.
The `download` method is otherwise exposed to non-numerical argument passing.

frgfm:
Closing this as the `valid_pct` argument has been removed since the train/test split is now set directly in the downloaded annotations as per #32. Feel free to reopen if you think we have missed something!

PR #28, opened 2019-10-17
feat: Split creation of path structure from actual download of images…

### LINKED ISSUE #26: [datasets] Split creation of path structure from actual download of images
## 🚀 Feature

Split creation of path structure and files that define train/test splitting from actual download of images

## Motivation

In order to use an external folder with images and preserve the test/train split defined in OpenFire without having to manually create the path structure.

blenzi:
… in OpenFire

- Add methods create_directories and check_images
- Rework method download
- Add option max_files
- Use max_files = 100 in test/test_datasets.py

Resolves: #26


pyronear/datasets/openfire.py
```diff
@@ -17,6 +17,18 @@
...
     """Wildfire image Dataset.
+    Creates the directory structure below and downloads the dataset or use the existing
+    files
+    
+    - root/
+    | - extract.json
+    | - OpenFire/
+    | | - processed/
+    | | |  - test.pl
+    | | |  - training.pl
```
frgfm: Little typo here: `.pt` extensions, not `.pl`

pyronear/datasets/openfire.py
```diff
@@ -42,13 +54,16 @@ class OpenFire(VisionDataset):
...
         self.train = train  # training set or test set
-
-        if download:
-            self.download(threads, valid_pct)
+        
+        if not (self._check_exists(train=True) and self._check_exists(train=False)):
+          self.create_directories(max_files=max_files)
+          if download:
+              self.download(threads)
+          self.check_images(valid_pct=valid_pct)
```
frgfm: Spacing issue here, one tab should be equal to 4 spaces in your editor (rather than 2 here most likely)

pyronear/datasets/openfire.py
```diff
@@ -42,13 +54,16 @@ class OpenFire(VisionDataset):
...
-                 download=False, threads=16, valid_pct=None):
+                 download=False, threads=16, valid_pct=0.2, max_files=-1):
         super(OpenFire, self).__init__(root, transform=transform,
                                     target_transform=target_transform)
         self.train = train  # training set or test set
-
-        if download:
-            self.download(threads, valid_pct)
+        
+        if not (self._check_exists(train=True) and self._check_exists(train=False)):
```
frgfm, line 62: Not sure to understand the logic behind this line.
The lines below will be executed only if test set exists already but not the training one?

pyronear/datasets/openfire.py
```diff
@@ -101,38 +116,55 @@ def _check_exists(self, train=True):
...
+            try:
+                self.annotations = json.load(f)[:max_files]
+            except TypeError:
+                raise ValueError("max_files must an int between -1 and len(dataset)")
 
-        # Download actual images
-        training_set, test_set = [], []
+        # Add the local filename to the annotations
+        for idx, a in enumerate(self.annotations):
+          a['fname'] = f"{idx:06}.{get_fname(a['url']).rpartition('.')[-1]}"
```
frgfm: Spacing issue here (should be 4 spaces rather than 2)

frgfm (changes requested):
While I understand the need to split:
- downloading the images
- verifying the images and saving the `.pt` extracts

There are two additional features that are being proposed here:

1. Separating hierarchy creation & download
I don't see any reason why one would be performed without the other?

2. Setting a limited number of files in extract
For actual public use of the library (non-testing purposes), limiting the download to a subtract of the whole might be troublesome if different users unexpectedly pass different values but expected to get the same dataset. Apart from unittests, is there any other use cases?

pyronear/datasets/openfire.py
```diff
@@ -101,38 +116,55 @@ def _check_exists(self, train=True):
...
             return self._root.joinpath(self._processed, self.test_file).is_file()
 
-    def download(self, threads=None, valid_pct=None):
-        """Download the OpenFire data if it doesn't exist in processed_folder already.
-
+    def create_directories(self, max_files=None):
+        """
+        Create the directory structure to hold the images and training_file/test_file.
+        Download the json file with annotations, add 'fname' with local filenames
+
```
frgfm: Trailing whitespace here according to Codacy

frgfm: There is a trailing whitespace here according to Codacy.

blenzi:
Thanks for the review!

> While I understand the need to split:
> 
>     * downloading the images
> 
>     * verifying the images and saving the `.pt` extracts
> 
> 
> There are two additional features that are being proposed here:
> 
>     1. Separating hierarchy creation & download
>        I don't see any reason why one would be performed without the other?

It was to prepare the next PR in which the user would be able to pass an existing image folder, to be linked to the one used by OpenFire.

Maybe I can ask differently: why this structure is needed ? Would it be enough to pass an image folder, which is downloaded if not given?

Then the code could store the .pt files used for training and testing in pytorch in a temporary location. This is fast and avoids the risk of a user removing the images while keeping the .pt files (I have done it accidentally already...).

And in the future, using other datasets could be done easily by passing another json file as argument as well. Maybe this is what you had in mind ?

> 
>     2. Setting a limited number of files in extract
>        For actual public use of the library (non-testing purposes), limiting the download to a subtract of the whole might be troublesome if different users unexpectedly pass different values but expected to get the same dataset. Apart from unittests, is there any other use cases?

Not that I can think of but I believe it is an important use case. Adding a comment that the parameter should only be changed for testing purposes would be enough in your opinion ?

frgfm:
> It was to prepare the next PR in which the user would be able to pass an existing image folder, to be linked to the one used by OpenFire.
I fully agree, but even if the user passes a dataset, the folder hierarchy will have to be verified, or am I missing something?
Probably having something similar to 
```
if not my_folder.is_folder():
    my_folder.mkdir(parents=True)
```
to leverage the use of `pathlib.Path` objects.

> Maybe I can ask differently: why this structure is needed ? Would it be enough to pass an image folder, which is downloaded if not given?

Good point, I thought about it more as a convention where all the libraries downloading datasets can share a same root (hence the default `./data` value), where all their datasets are stored. PyTorch does it both for datasets and models. 
The user can easily free some memory if she/he wants to delete the downloaded data at once (or move it for that matter). And after that, expectedly, you have to split the data within the root depending on its project/source.

> Then the code could store the .pt files used for training and testing in pytorch in a temporary location. This is fast and avoids the risk of a user removing the images while keeping the .pt files (I have done it accidentally already...).

Good idea, fully agree with you on that one.

> And in the future, using other datasets could be done easily by passing another json file as argument as well. Maybe this is what you had in mind ?

Yes, the core idea is that the  ̀OpenFire` dataset will either be split into `OpenFireBin` `OpenFireMulti`, etc for the different tasks, or take an argument `task` (preferably). Based on the value of `task` it will either fetch the  ̀.json` from a different URL (difference in annotations), or download everything at once, and return the appropriate version of the annotated dataset (my suggestion).

> Not that I can think of but I believe it is an important use case. Adding a comment that the parameter should only be changed for testing purposes would be enough in your opinion?

My point is more that exposing methods/functions for users while the feature is only used by developers (actually code maintainers, not necessarily developers) and confusing for users, might not be our best option here. Ideally, the operation of limiting the extract will be only available and performed within the test, but I agree it's not easily done. A good lead might be to take a look at https://github.com/pytorch/vision/blob/master/test/test_datasets.py#L84-L91

That being said, if the point is about speeding up the CI job, in comparison, we are still at a very reasonable duration compared to `torchvision` or other similar libraries.

pyronear/datasets/openfire.py
blenzi: They will be skipped if both exist. Maybe you missed the () ?
(N.B.: this will disappear if we create the .pt files each time)

pyronear/datasets/openfire.py
blenzi: Thanks, fixed

pyronear/datasets/openfire.py
blenzi: Fixed

pyronear/datasets/openfire.py
blenzi: Fixed

pyronear/datasets/openfire.py
blenzi: Fixed (codacy is annoying, this is just documentation)

pyronear/datasets/openfire.py
frgfm: Indeed, my bad. Better to go for  `not A or not B` (it will stop earlier if the first condition is met, whereas in your suggestion both will be evaluated)

blenzi: Aesthetically I agree but I believe both are equivalent. The following does not execute the 2nd statement:

```
import time
if not (False and time.sleep(5)): pass
```

frgfm: I agree but I was referring to execution time, but I just tested it and it doesn't seem to be doing much difference indeed so that's alright!

pyronear/datasets/openfire.py
```diff
@@ -101,38 +116,55 @@ def _check_exists(self, train=True):
         else:
             return self._root.joinpath(self._processed, self.test_file).is_file()
 
-    def download(self, threads=None, valid_pct=None):
-        """Download the OpenFire data if it doesn't exist in processed_folder already.
+    def create_directories(self, max_files=None):
```
frgfm: default value of `max_files` differs from line 57, is it normal?

blenzi: Thanks for catching it. It should be None there as well, otherwise the last file would be skipped (it was a bug on the original implementation)

frgfm: Also, perhaps rename it to `num_samples` for a less confusing variable naming

pyronear/datasets/openfire.py
```diff
@@ -101,38 +116,55 @@ def _check_exists(self, train=True):
...
         self._root.joinpath(self._processed).mkdir(parents=True, exist_ok=True)
 
         # Download annotations
         download_url(self.url, self._root.joinpath(self._raw), filename=self.url.rpartition('/')[-1], verbose=False)
         with open(self._root.joinpath(self._raw, self.url.rpartition('/')[-1]), 'rb') as f:
-            annotations = json.load(f)
+            try:
+                self.annotations = json.load(f)[:max_files]
+            except TypeError:
+                raise ValueError("max_files must an int between -1 and len(dataset)")
```
frgfm: You put a `TypeError` in except but raise a `ValueError`.
I would rather go for something like:
```
self.annotations = json.load(f)
if max_files is not None:
    self.annotations = self.annotations[:max_files]
```
It will throw the error by itself if indexing type/value isn't correct

However, I don't get why negative values are allowed considering the feature is limiting data samples (not indexing them)

pyronear/datasets/openfire.py
```diff
@@ -42,13 +54,16 @@ class OpenFire(VisionDataset):
     seed = 42
 
     def __init__(self, root, train=True, transform=None, target_transform=None,
-                 download=False, threads=16, valid_pct=None):
+                 download=False, threads=16, valid_pct=0.2, max_files=-1):
```
frgfm: default value of `max_files` means that the extract will not take the last sample. I'm not sure it is expected as a behaviour! The default value should produce the same output as before the feature

pyronear/datasets/openfire.py
```diff
@@ -17,6 +17,18 @@
...
+
+    - root/
+    | - extract.json
+    | - OpenFire/
+    | | - processed/
+    | | |  - test.pt
+    | | |  - training.pt
+    | | - raw/
+    | | | - images/
+
```
frgfm, line 31: No need for double line break here

pyronear/datasets/openfire.py
```diff
@@ -101,38 +116,54 @@ def _check_exists(self, train=True):
...
         self._root.joinpath(self._processed).mkdir(parents=True, exist_ok=True)
 
         # Download annotations
         download_url(self.url, self._root.joinpath(self._raw), filename=self.url.rpartition('/')[-1], verbose=False)
         with open(self._root.joinpath(self._raw, self.url.rpartition('/')[-1]), 'rb') as f:
-            annotations = json.load(f)
+            try:
+                self.annotations = json.load(f)[:num_samples]
+            except TypeError:
+                raise ValueError("num_samples must an int between -1 and len(dataset)")
```
frgfm, line 136: Better to go for
```
self.annotations = json.load(f)
if num_samples is not None:
    self.annotations = self.annotations[:num_samples]
```
It will throw the appropriate error by itself if the indexing is incorrect

pyronear/datasets/openfire.py
```diff
@@ -101,38 +116,54 @@ def _check_exists(self, train=True):
...
 
-        # Download actual images
-        training_set, test_set = [], []
+        # Add the local filename to the annotations
+        for idx, a in enumerate(self.annotations):
+            a['fname'] = f"{idx:06}.{get_fname(a['url']).rpartition('.')[-1]}"
```
frgfm, line 140:
```
self.annotations[idx]['fname'] = ...
```
would be more appropriate for clear inplace modification of class attribute

pyronear/datasets/openfire.py
```diff
@@ -42,13 +54,16 @@ class OpenFire(VisionDataset):
...
                                     target_transform=target_transform)
         self.train = train  # training set or test set
 
-        if download:
-            self.download(threads, valid_pct)
+        if not (self._check_exists(train=True) and self._check_exists(train=False)):
+            self.create_directories(num_samples=num_samples)
+            if download:
+                self.download(threads)
+            self.check_images(valid_pct=valid_pct)
```
frgfm, line 66: We have to think a bit much about this:
if you provide a different value of  `valid_pct` with `download=True`, currently, the condition of  `self._check_exists` will not be verified. So it will use extract that was created with a different value of `valid_pct`

pyronear/datasets/openfire.py
blenzi: Done

frgfm (changes requested):
A few changes to make for the feature to be compatible and have a good expected behaviour.
The feature proposes to split the following operations:
- folder hierarchy creation
- sample downloading
- on-the-fly extract saving with custom train/valid split

Here would be the wanted behaviour:
- if  `download=False`: the class will try to use existing saved extracts (throwing an error otherwise like previously)
- if `download=True`: 
1. folder hierarchy will be created if it does not exist yet.
2. the JSON extract will be download if it does not exist yet
3. the actual images will be downloaded (if `num_samples` is provided, only the first `num_samples` will be downloaded)
4. samples will be verified (throwing a warning if the specified amount of samples has not been downloaded or is not available)

- Whatever  the value of `download`:
5. `valid_pct` of total available samples will used for validation and the rest for training

If `download=True`:
- steps 1, 2 can be skipped
- step 3 cannot (internally it will skip download if the file already exists)
- step 4 cannot
Extract saving at this point might be irrelevant
- step 5 is always executed

blenzi:
Thanks for the in-depth review of the logic. While I agree with most of the comments, I believe we are missing the reasons why this issue/PR was open:

- Downloading the entire dataset is not fast and is essentially never succesful. The user is forced to either manually create a folder structure that was not explicitly documented or interrupt the execution and copy 3 GB into a folder inside a nested structure. This is not convenient, in particular when using a system that requires starting from scratch like google colab.

- Without limiting the number of files, no quick tests of new functionalities or new training methods are possible (at least easily).

- The class was designed to be interfaced with torch directly while working with fastai requires quite a few workarounds. The train/test split and the .pt files are not particularly relevant.

The suggestion would be to add optional arguments:

- num_samples (already implemented) 
- image_folder, to be linked to OpenFire/raw/images (instead of copying)
- Allow for train=None, 'all' or change the argument name, in order to get the full dataset
- Add method to_ImageList for easy use with fastai

Also:

>     * Whatever  the value of `download`:
> 
> 
>     1. `valid_pct` of total available samples will used for validation and the rest for training

The current behaviour is to use the existing train/test splitting if download is False. Is this what is wanted ?

Maybe it's easier to drop this PR and restart from a fresh one if you agree with the points above

frgfm:
I agree @blenzi 
Let's decide the design specs on the issue #26 and work on a clean PR for it!

PR #30, opened 2019-10-20
fix: Fixed training scripts

frgfm:
This PR aims at reducing the discrepancies between training run with fastai and bare pytorch. The later was running by default with fully unfrozen model while the first was only unfreezing the fully connected layers.

Also added checkpoint resuming option on both scripts, and a binary classification option on the pytorch script

### DIFF
--- a/references/classification/fastai/train.py
+++ b/references/classification/fastai/train.py
@@ -50,6 +50,8 @@ def main(args):
                                  ps=args.dropout_prob,
                                  concat_pool=args.concat_pool,
                                  metrics=vision.error_rate)
+    if args.resume:
+        learner.load(args.resume)
     if args.unfreeze:
         learner.unfreeze()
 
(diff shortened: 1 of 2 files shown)


references/classification/fastai/train.py
```diff
@@ -71,21 +75,20 @@ def main(args):
...
                         dest='weight_decay')
     parser.add_argument('--div-factor', default=25., type=float, help='div factor of OneCycle policy')
     parser.add_argument('--checkpoint', default='checkpoint', type=str, help='name of output file')
-    parser.add_argument(
-        "--pretrained",
-        dest="pretrained",
-        help="Use pre-trained models from the modelzoo",
-        action="store_true",
-    )
+    parser.add_argument('--resume', default=None, help='resume from checkpoint')
```
blenzi: It's a detail but the way the help is written, the use may think it is a yes/no while the user has to provide the name of the checkpoint if I understand well.

blenzi (approved):
Maybe you know this: in the future, in case you want to show the default arguments, you can use:
https://docs.python.org/3/library/argparse.html#argparse.ArgumentDefaultsHelpFormatter

PR #32, opened 2019-10-21
OpenFire extract update

frgfm:
This PR updates the `OpenFire` extract to add two information for each sample:
- `is_test`: whether the sample belongs to the validation/test subset or training set
- `name`: expected filename

Any feedback is welcome!

### DIFF
--- a/pyronear/datasets/openfire.py
+++ b/pyronear/datasets/openfire.py
@@ -34,7 +34,7 @@ class OpenFire(VisionDataset):
             for downloading the dataset.
     """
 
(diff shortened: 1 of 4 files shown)


blenzi (dismissed):

blenzi (approved):

x0s:
Why did you change the structure of the JSON ? Where is it documented ?
Who decides what is 'is_test' or not ? 

This PR seems to suppress the ability for the user to choose a validation set, hence a validation strategy.

frgfm:
Regarding the changes, I suggest you check the discussion with @blenzi in #28 and #26 

For the JSON, keys were added for each sample but nothing has been deleted. The structure of each dict is available at its URL but basically:
- url: URL of the image
- name: name of the image file that will be saved
- target: annotation
- is_test: whether the sample is in test/valid set or in the training set

PR #33, opened 2019-10-22
feat: add argument num_samples to OpenFire

blenzi:
Add the possibility to limit the number of samples to be downloaded, as discussed in #26. The default value is None, meaning all samples will be downloaded as before. Limited to 200 in test_datasets.py

### DIFF
--- a/pyronear/datasets/openfire.py
+++ b/pyronear/datasets/openfire.py
@@ -31,6 +31,7 @@ class OpenFire(VisionDataset):
             downloaded again.
         threads (int, optional): If download is set to True, use this amount of threads
             for downloading the dataset.
+        num_samples (int, optional): Number of samples to download (all by default)
     """
 
     url = 'https://gist.githubusercontent.com/frgfm/f53b4f53a1b2dc3bb4f18c006a32ec0d/raw/c0351134e333710c6ce0c631af5198e109ed7a92/openfire_binary.json'
@@ -38,12 +39,12 @@ class OpenFire(VisionDataset):
     test_file = 'test.pt'
     classes = [False, True]
 
-    def __init__(self, root, train=True, download=False, threads=16, **kwargs):
+    def __init__(self, root, train=True, download=False, threads=16, num_samples=None, **kwargs):
         super(OpenFire, self).__init__(root, **kwargs)
         self.train = train  # training set or test set
 
         if download:
-            self.download(threads)
+            self.download(threads, num_samples)
 
         if not self._check_exists(train):
             raise RuntimeError('Dataset not found.' +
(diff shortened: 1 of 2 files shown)


pyronear/datasets/openfire.py
```diff
@@ -112,7 +114,7 @@ def download(self, threads=None):
         # Download annotations
         download_url(self.url, self._root.joinpath(self._raw), filename=self.url.rpartition('/')[-1], verbose=False)
         with open(self._root.joinpath(self._raw, self.url.rpartition('/')[-1]), 'rb') as f:
-            annotations = json.load(f)
+            annotations = json.load(f)[:num_samples]
```
frgfm, line 117: I'm not sure, but shouldn't it be:
```
annotations = json.load(f)
if num_samples is not None:
    annotations = annotations[:num_samples]
```

Otherwise, slicing with None will result in error

test/test_datasets.py
```diff
@@ -47,8 +47,8 @@ def test_openfire(self):
         with Path(tempfile.TemporaryDirectory().name) as root:
 
             # Working case
-            train_set = datasets.OpenFire(root=root, train=True, download=True)
-            test_set = datasets.OpenFire(root=root, train=False, download=True)
+            train_set = datasets.OpenFire(root=root, train=True, download=True, num_samples=200)
+            test_set = datasets.OpenFire(root=root, train=False, download=True, num_samples=200)
```
frgfm: Better to add `num_samples = 200` as a test variable above line 47 and pass it to the appropriate functions.
So that if we want to increase/decrease that value, only one line has to be changed

test/test_datasets.py
```diff
@@ -58,12 +58,12 @@ def test_openfire(self):
             self.assertTrue(all(sample['path'].name.rpartition('.')[-1] in ['jpg', 'jpeg', 'png', 'gif']
                                 for sample in test_set.data))
 
-            # Check against number of samples in extract
+            # Check against number of samples in extract (limit to 200)
```
frgfm: Small update to reflect my previous comment would be nice!

test/test_datasets.py
```diff
@@ -58,12 +58,12 @@ def test_openfire(self):
...
             datasets.utils.download_url(train_set.url, root, filename='extract.json', verbose=False)
             with open(root.joinpath('extract.json'), 'rb') as f:
-                extract = json.load(f)
+                extract = json.load(f)[:200]
```
frgfm: Same here

frgfm (changes requested):

pyronear/datasets/openfire.py
blenzi: None is a perfectly valid argument for a slice and keeps all the list
```
In [4]: range(10)[:None]
Out[4]: range(0, 10)
```

test/test_datasets.py
blenzi: Agreed!

test/test_datasets.py
blenzi: Done

test/test_datasets.py
blenzi: Done

pyronear/datasets/openfire.py
frgfm: My bad then, thanks for your response!

frgfm (approved):
PR adds an optional parameter to limit the number of samples that would be downloaded by `OpenFire.download`.

PR #34, opened 2019-10-22
feat: Add argument img_folder to OpenFire

blenzi:
Following the discussion in #26, add argument img_folder (str or Path) to OpenFire. If not given, the default  `<root>/OpenFire/images` is used. The rest of the folder structure is kept for the moment.

test_datasets.py now tests its type and also if the download succeeded approximately: up to 10 samples can be missing.


pyronear/datasets/openfire.py
```diff
@@ -32,16 +32,19 @@ class OpenFire(VisionDataset):
...
+        img_folder (str or Path, optional): Location of image folder. Default: <root>/OpenFire/images
     """
 
     url = 'https://gist.githubusercontent.com/frgfm/f53b4f53a1b2dc3bb4f18c006a32ec0d/raw/c0351134e333710c6ce0c631af5198e109ed7a92/openfire_binary.json'
     training_file = 'training.pt'
     test_file = 'test.pt'
     classes = [False, True]
 
-    def __init__(self, root, train=True, download=False, threads=16, num_samples=None, **kwargs):
+    def __init__(self, root, train=True, download=False, threads=16, num_samples=None, img_folder=None, **kwargs):
```
frgfm: Would you mind going to next line after `num_samples` term please?
I'll open a PR with pep8 verification (including max_line_length=120) and this line is longer apparently

pyronear/datasets/openfire.py
```diff
@@ -32,16 +32,19 @@ class OpenFire(VisionDataset):
...
         super(OpenFire, self).__init__(root, **kwargs)
         self.train = train  # training set or test set
+        self.img_folder = self._root.joinpath(self.__class__.__name__, 'images') \
+          if img_folder is None else Path(img_folder)
```
frgfm: Better go for
```
if img_folder is None:
    ....
else:
     ....
```
which would remove the need for line continuation

test/test_datasets.py
```diff
@@ -44,12 +44,17 @@ def test_downloadurls(self):
 
     def test_openfire(self):
         num_samples = 200
+        img_folder = tempfile.TemporaryDirectory().name # N.B.: different from root
 
         with Path(tempfile.TemporaryDirectory().name) as root:
```
frgfm: Perhaps this
```
with Path(tempfile.TemporaryDirectory().name) as root, Path(tempfile.TemporaryDirectory().name) as img_folder:
    ....
``` 
would be cleaner?

frgfm: Actually I just checked, and we can do:
```
with tempfile.TemporaryDirectory() as root, tempfile.TemporaryDirectory() as img_folder:
    ...
```
and only change the `extract.json` line with  `Path(root).joinpath` rather than `root.joinpath`

frgfm (changes requested):
Great PR!
Just a little edit for good pep8 compliance if you don't mind

pyronear/datasets/openfire.py
```diff
@@ -118,17 +124,16 @@ def download(self, threads=None, num_samples=None):
...
         entries = [(a['url'], a['name']) for idx, a in enumerate(annotations)]
         # Use multiple threads to speed up download
-        download_urls(entries, img_folder, threads=threads)
+        download_urls(entries, self.img_folder, threads=threads)
         # Verify downloads
         for idx, annotation in enumerate(annotations):
-            img_path = self._raw.joinpath('images', entries[idx][1])
-            if self._root.joinpath(img_path).is_file():
+            img_path = self.img_folder.joinpath(entries[idx][1])
+            if img_path.is_file():
```
frgfm: Are you positive it works when iterating over the dataset as a dataloader?
Initially `img_path` was only starting under `root` because once it's stored in the dictionary on line 140, the image will then be loaded as `root.joinpath(img_path)` if I'm not mistaking.
```
loader = torch.utils.data.DataLoader(dataset, batch_size=1, num_workers=4)
for x, target in loader:
    y = target
```
Does this work on your end? (or the pytorch training reference script for that matter)


Perhaps we should add a unittest with data loader to check this in CI

blenzi: Starting from scratch this works:
```
mkdir -p /tmp/data/OpenFire
ln -s ~/Workarea/PyroNear/data/OpenFire/images /tmp/data/OpenFire/
python references/classification/torch/train.py --device cpu --workers 0 --epochs 1 --data-path /tmp/data
```

`ls -R /tmp/data` gives:

```
OpenFire

/tmp/data/OpenFire:
images    processed raw

/tmp/data/OpenFire/processed:
test.pt     training.pt

/tmp/data/OpenFire/raw:
openfire_binary.json
```

I am not familiar with those classes but:
- iterating over the dataset works
- each element is a tuple (PIL.Image.Image, target)
- the test `is_file()` in the lines above succeeds since `train.pt` and `test.pt` are created as before.
- The one you suggest doesn't because it is an image, so: 

```
import torch, pyronear
dataset = pyronear.datasets.OpenFire('/tmp/data')
loader = torch.utils.data.DataLoader(dataset, batch_size=1, num_workers=4)
for x, target in loader:
    y = target

TypeError: default_collate: batch must contain tensors, numpy arrays, numbers, dicts or lists; found <class 'PIL.Image.Image'>
```

And yes, I totally agree with adding tests for dataloader, etc

frgfm: Perfect, I just wanted to make sure we don't compromise the training!
For the collate function, it requires to resize all images to the same size, but if the training is running then it's all god
Thank you very much!

frgfm (dismissed):
Very useful PR to symlink image folder located elsewhere for training!
A few extra tests will have to be added to unittests to monitor further developments of OpenFire but this one is working as is.

frgfm:
@blenzi Just checking, is this still on? Or was it covered by another PR recently?
If no longer relevant, would you mind closing the PR please?

frgfm (dismissed):
Thanks for clearing the PR! Looks good to me

x0s:
@blenzi : Do you need any help to resolve the conflicts ?

blenzi:
> @blenzi : Do you need any help to resolve the conflicts ?

Thanks @x0s. I didn't check it yet and will not be able to do in the next day or two. If you want, please go ahead!

frgfm:
Mind trying to solve the conflicts @blenzi :pray: ? (the scripts were moved, just need some fixing)

frgfm (approved):
Thanks for fixing the PR!

PR #35, opened 2019-10-23
style: Added pep8 verification to CI

frgfm:
Adds code style verification according to pep8. The corresponding job in CI is named `pkg-style` and will point out where collaborators can improve their pep8 code compliance.

Any feedback is welcome!

_Note: errors and warnings that are being ignored in the `.flake8` file may be updated in the future_

### DIFF
--- a/pyronear/datasets/openfire.py
+++ b/pyronear/datasets/openfire.py
@@ -34,7 +34,8 @@ class OpenFire(VisionDataset):
         num_samples (int, optional): Number of samples to download (all by default)
     """
 
-    url = 'https://gist.githubusercontent.com/frgfm/f53b4f53a1b2dc3bb4f18c006a32ec0d/raw/c0351134e333710c6ce0c631af5198e109ed7a92/openfire_binary.json'
+    url = ('https://gist.githubusercontent.com/frgfm/f53b4f53a1b2dc3bb4f18c006a32ec0d/raw/'
+           'c0351134e333710c6ce0c631af5198e109ed7a92/openfire_binary.json')
     training_file = 'training.pt'
     test_file = 'test.pt'
     classes = [False, True]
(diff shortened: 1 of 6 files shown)


pyronear/datasets/openfire.py
```diff
@@ -34,7 +34,8 @@ class OpenFire(VisionDataset):
         num_samples (int, optional): Number of samples to download (all by default)
     """
 
-    url = 'https://gist.githubusercontent.com/frgfm/f53b4f53a1b2dc3bb4f18c006a32ec0d/raw/c0351134e333710c6ce0c631af5198e109ed7a92/openfire_binary.json'
+    url = ('https://gist.githubusercontent.com/frgfm/f53b4f53a1b2dc3bb4f18c006a32ec0d/raw/'
```
blenzi: Breaking the url in multiple lines is the recommendation ?

frgfm: flake8 seems to pick it as a too long line, but I just found that there might be a trick: https://stackoverflow.com/questions/10739843/how-should-i-format-a-long-url-in-a-python-comment-and-still-be-pep8-compliant

Keeping the line as is and add `# noqa` comment at the end of the line so that flake doesn't pick it. I'll try this

blenzi (approved):

pyronear/datasets/openfire.py
x0s: Yep, but that's better practice to specify the check you want to avoid so no other mistake might be merged. `# noqa: E501` 
(https://lintlyci.github.io/Flake8Rules/rules/E501.html)

frgfm: Thanks for the suggestion! Would you mind opening a PR for the fix @x0s?
It's indeed a much better option.

Issue #36, opened 2019-10-23
[ci] Unpickling error for pretrained parameter loading


frgfm:
## 🐛 Bug

The `pretrained=True` option on models cannot be tested by CI, since passing it produces a `pickle` error that cannot be reproduced locally.

## To Reproduce

1. Running the `pkg-test` CI job on this [commit](https://github.com/frgfm/PyroNear/commit/079a2ee880cbdf37774d423a53bb4d2b471170e4)

which produces:
```
Traceback (most recent call last):
  File "/home/circleci/repo/test/test_models.py", line 22, in test_resnet
    model = models.resnet(depth=18, pretrained=True, bin_classif=True).eval()
  File "/home/circleci/repo/pyronear/models/resnet.py", line 49, in resnet
    progress=progress)
  File "/home/circleci/repo/venv/lib/python3.6/site-packages/torch/hub.py", line 499, in load_state_dict_from_url
    return torch.load(cached_file, map_location=map_location)
  File "/home/circleci/repo/venv/lib/python3.6/site-packages/torch/serialization.py", line 426, in load
    return _load(f, map_location, pickle_module, **pickle_load_args)
  File "/home/circleci/repo/venv/lib/python3.6/site-packages/torch/serialization.py", line 603, in _load
    magic_number = pickle_module.load(f, **pickle_load_args)
_pickle.UnpicklingError: invalid load key, 'Y'.
```

## Expected behavior

State dictionary loading should not produce errors.

## Environment

Please copy and paste the output from our
[environment collection script](https://raw.githubusercontent.com/frgfm/PyroNear/master/pyronear/utils/collect_env.py)
(or fill out the checklist below manually).

You can get the script and run it with:
```
wget https://raw.githubusercontent.com/frgfm/PyroNear/master/pyronear/utils/collect_env.py
# For security purposes, please check the contents of collect_env.py before running it.
python collect_env.py
```

 - PyTorch Version: 1.3.0
 - Torchvision Version: 0.4.1a0+d94043a
 - OS: Ubuntu 18.04
 - How you installed PyTorch: `conda`
 - Python version: 3.7
 - CUDA/cuDNN version: CUDA 10.1.168 / cuDNN 7.6.2
 - GPU models and configuration: GeForce GTX 1050 (driver: 430.50)

frgfm:
This issue seems to have been fixed a while back, I'm not able to reproduce it anymore.
Closing for now, feel free to reopen if you encounter it!

PR #37, opened 2019-10-23
feat: Added the models module

frgfm:
This PR introduces the `models` module which:
- defines resnet architectures for binary classification (on `OpenFire`)
- allows pretrained parameters download when `pretrained=True`

Tests and documentation have been updated accordingly.
_Note: the only available pretrained model for now is a `resnet18`. Other variations will need to be added._

Any feedback is welcome!

### DIFF
--- a/pyronear/__init__.py
+++ b/pyronear/__init__.py
@@ -1,3 +1,4 @@
 from pyronear import datasets
+from pyronear import models
 from pyronear import utils
 from .version import __version__
--- a/pyronear/models/__init__.py
+++ b/pyronear/models/__init__.py
@@ -1,0 +1,1 @@
+from .resnet import *
--- a/pyronear/models/__init__.py
+++ b/pyronear/models/__init__.py
@@ -1,1 +1,1 @@
-from .resnet import *
+from .resnet import resnet
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -21,8 +21,8 @@ import sphinx_rtd_theme
 
 master_doc = 'index'
 project = 'pyronear'
-copyright = '2019, PyroNear Contibutors'
-author = 'PyroNear Contibutors'
+copyright = '2019, PyroNear Contributors'
+author = 'PyroNear Contributors'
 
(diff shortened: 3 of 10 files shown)


pyronear/__init__.py
```diff
@@ -1,3 +1,4 @@
 from pyronear import datasets
+from pyronear import models
 from pyronear import utils
 from .version import __version__
```
blenzi, line 4: This should not fail your pep8 checks ?

pyronear/models/__init__.py
```diff
@@ -0,0 +1 @@
+from .resnet import *
```
blenzi: Do you need `import *`  ? :)

pyronear/__init__.py
frgfm: How so?

blenzi (commented):
Nice! Is there a way to check the functionality of using the model(s) ?

PS: it's not related to the models but good that you changed the name of the test classes

pyronear/__init__.py
blenzi: No newline at the end of file

frgfm: Oh, the branch was created before pep8 was introduced. So when merging into master (where that end of line has been added), it will be fine.

frgfm:
> Nice! Is there a way to check the functionality of using the model(s) ?
> 
> PS: it's not related to the models but good that you changed the name of the test classes

Not sure what you mean by that but  `test/test_models.py` is here for checking expected behaviours

blenzi:
> > Nice! Is there a way to check the functionality of using the model(s) ?
> > PS: it's not related to the models but good that you changed the name of the test classes
> 
> Not sure what you mean by that but `test/test_models.py` is here for checking expected behaviours

I mean checking if the model is properly loaded for instance

blenzi (approved):

frgfm:
> 
> I mean checking if the model is properly loaded for instance

CI has an issue with ̀pretrained=True` so I removed it. But I opened an issue to track this #36

PR #41, opened 2019-10-25
feat: Unified model creation for CNN

frgfm:
The PR proposes multiple features:
- new module `pyronear.models.utils`: tools for model creation which allows easy standardization of cnn models.
- update `pyronear.models`:  default model's head has been switched to the implementation of fastai, and pretrained parameters' urls have been updated. Variations of resnet, densenets and mobilenet have been added.
- update `references/classification`: both scripts now allow binary classification training and architectures are identical with default arguments.
- created `pyronear.nn`: added concatenated pooling module

Any feedback is welcome!

### DIFF
--- a/pyronear/models/densenet.py
+++ b/pyronear/models/densenet.py
@@ -34,7 +34,8 @@ def _update_state_dict(state_dict):
 
 
 def _densenet(arch, growth_rate, block_config, num_init_features, pretrained=False,
-              progress=True, imagenet_pretrained=False, num_classes=1, **kwargs):
+              progress=True, imagenet_pretrained=False, num_classes=1, lin_features=None,
+              dropout_prob=0.5, bn_final=False, concat_pool=True, **kwargs):
     r"""Instantiate a DenseNet model for image classification from
     `"Densely Connected Convolutional Networks" <https://arxiv.org/pdf/1608.06993.pdf>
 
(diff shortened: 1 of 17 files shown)


pyronear/models/densenet.py
```diff
@@ -0,0 +1,150 @@
...
+import re
+from torchvision.models.densenet import DenseNet, model_urls as imagenet_urls
+from torchvision.models.utils import load_state_dict_from_url
+from .utils import cnn_model
+
+__all__ = ['densenet121', 'densenet169', 'densenet201', 'densenet161']
+
+
+model_urls = {
+    'densenet121': 'https://srv-file7.gofile.io/download/XqHLBB/densenet121-binary-classification.pth'
```
blenzi, line 13: Is it normal that there is only a URL for densenet121 and not the others ?

references/classification/fastai/train.py
```diff
@@ -37,61 +75,94 @@ def main(args):
...
 
 if __name__ == "__main__":
     import argparse
     parser = argparse.ArgumentParser(description='PyroNear Classification Training with Fastai')
+    # Input / Output
     parser.add_argument('--data-path', default='./data', help='dataset')
+    parser.add_argument('--checkpoint', default='checkpoint', type=str, help='name of output file')
+    parser.add_argument('--resume', default=None, help='checkpoint name to resume from (default: None)')
+    # Architecture
     parser.add_argument('--model', default='resnet18', type=str, help='model')
```
blenzi: Since now there are many possibilities, you could add the default to the helper (explicitly or using ArgumentDefaultsHelpFormatter for all arguments)

blenzi (commented):
Very nice @frgfm ! The parsers for fastai and torch look quite similar from first sight. Maybe it would be good (at some point) to create a common structure a simply have the different bits in each of the scripts ?

pyronear/models/densenet.py
frgfm: It's just that I didn't have time to train the others!
But the same goes for all `resnext` and `wide_resnet` unfortunately. I thought we would deal with that in the next release (since it takes at least 40-45min per model and there are quite a lot of them), what do you think?

blenzi (approved):

pyronear/models/densenet.py
blenzi: If it does not affect the normal usage and the user will get an understandable error when trying to load a pretrained model (if I understood the idea), I think it should be fine

PR #46, opened 2019-12-03
Add wildfire auto fire labeler

x0s:
This PR aims at adding a auto fire labeling feature to the WildFire dataset.

The new `FireLabeler` class tries to group the videos representing the same fire. Labeling is based on words in descriptions of the videos. It looks for commonalities through a sliding window with a size one can vary representing the depth of the search.

This PR also create a `wildfire` directory in `datasets` module. I suggest to group in this directory the growing number of tools we are currently developing to preprocess this dataset.

Here is an example:

```python
df = pd.read_csv("WildFire.csv", index_col=0)
fire_labeler = FireLabeler(df, window_size=30)
fire_labeler.run()
df_on_fire = fire_labeler.get_new_dataframe(column_name='fire_id')

df_on_fire_short = (fire_labeler.reset(window_size=6)
                                .run()
                                .get_dataframe(column_name='fire_id'))
```
For more example, one can browse unit and integration tests.
Any feedback is welcome.

### DIFF
--- a/pyronear/datasets/wildfire/fire_labeler.py
+++ b/pyronear/datasets/wildfire/fire_labeler.py
@@ -1,8 +1,9 @@
 import re
-
 from itertools import combinations
+
 import numpy as np
 import pandas as pd
+
 
 class FireLabeler:
     """Automatically labelize WildFire dataset based on video descriptions
@@ -61,7 +62,7 @@ class FireLabeler:
         self.window_size = window_size or self.window_size
 
         self.n_videos_total = self.df.shape[0]
-        self.n_windows = self.n_videos_total - self.window_size + 1 # n_windows + windows_size < n_videos_total
+        self.n_windows = self.n_videos_total - self.window_size + 1  # n_windows + windows_size < n_videos_total
 
         # Store new column for fire ids starting at 0 (-1 for unassigned)
         self.fire_ids = np.full((self.n_videos_total), -1)
@@ -79,10 +80,10 @@ class FireLabeler:
 
         # sliding iterator over the video indexes. Example with window_size=4:
         # [[0, 1, 2, 3], [1, 2, 3, 4], [2, 3, 4, 5], ...]
-        window_idx_it = (range(start, start+self.window_size) for start in range(self.n_windows))
+        window_idx_it = (range(start, start + self.window_size) for start in range(self.n_windows))
 
-        current_fire_id = 0 # start grouping feu id at 0
-        for window_idx in window_idx_it: # for every window of videos
+        current_fire_id = 0  # start grouping feu id at 0
+        for window_idx in window_idx_it:  # for every window of videos
 
             # dict with {id: description(string)}
             id_to_descriptions = dict(zip(window_idx, self.df.loc[window_idx, 'description']))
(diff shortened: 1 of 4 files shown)


pyronear/datasets/wildfire/fire_labeler.py
```diff
@@ -0,0 +1,150 @@
...
+        return self
+
+    def run(self):
+        """ Run the labelisation of the fire depending on the descriptions
+
+        For every combination of descriptions(strings) in every sliding window
+        guess if they belong to same fire (fire_id)
+
+        Note: algo complexity can be improved (ex: by memoizing sliding combinations).
+        But, for now, processing is fast enough(<1min) when used with large window-size<100
```
blenzi, line 77: maybe I missed the logic but why a window size is needed and why it uses windows of fixed size for consecutive entries? Couldn't it compare pairs of descriptions and keep only the different "groups" ?

blenzi (commented):
Very nice @x0s ! Just one comment for the record: this will clearly help grouping the scenes with fire. On the other hand I believe there are many scenes from the same camera with no fire spread across multiple videos. And unfortunately this is not present in the descriptions.

pyronear/datasets/wildfire/fire_labeler.py
x0s: Thanks,
You are absolutely right. It _tries_ to group according to same fire so we minimize leakage *in time* between datasets splits(train/val/test). Surely, same camera lookouts may be spread among each split; to a certain extent, the trained model may not generalized very well, but practically it should be fined-tuned or retrained with data collected from the field. This should thwart generalization gap.

x0s: Thanks for your question.Concerning the window size:
- Yes it is needed, because we are looking for pairs not only in consecutive entries (it would be `windows_size` of 2) but also entries quite far away. You can browse the tests to see how it can benefit when dealing with long dependency. For instance Matching `KingFire` and `King Fire` requires to have a pair containing these two, but this is not happening consecutively.
- Comparing _every pair_ of descriptions would be too costly: complexity grows to the square of the count of videos you take into account:
image
So for a window of `n = 100 videos`, it roughly makes  *5000 iterations*,
but for a window of `n = 1000 videos` (all), it will require *500 000 iterations*, so 100 times more. In this case it will require more optimization, and I think this is an overkill for our problem right now, because we are trying to group videos representing same fire,videos that were uploaded, a priori, in the same order they were recorded.

Does this help?

blenzi: Yeah, I thought that indeed you tried to make use of the sequence. But I did not mean to compare every pair in practice. The idea was something inspired in merge sort: you compare two descriptions and if they match, you keep one as a representative of the group, otherwise you keep both. 

It can end up as n**2 if you compare all pairs and there are no similar ones but you can also compare only the adjacent ones to reduce the size. 

Just an idea. if it works already, then fine.

x0s: Thanks for the suggestion. You're right, when designing the algorithm, I wanted to be able to use the proximity of the descriptions as a reinforcement factor to detect matching. In the end, I did not use this because I think this is good enough. Your suggestion is also interesting but it would be more difficult to make use of the sequence. For now and for this dataset, I don't see what it could practically improve.

blenzi (approved):

PR #48, opened 2020-03-23
Add frame extractor for wildfire

x0s:
This PR aims at importing the frame extraction into the repository used for training models.

Until now, Frame Extraction was mingled with annotation process. This PR improves encapsulation by modularizing the frame extraction. We will be able to choose different frame extraction strategy for **training** (not only annotating). This is important feature because we will build/train our models on top of what we show to them (the frames)

This is an ongoing PR. It is important to no have duplicate code, so I consider importing this new class into the other repository.

I provide some graphs illustrating the current workflow, the new one I suggest (big picture and isolated).

This PR does not primarily aim to change behaviour but when covering with tests, I discover an issue that should be fixed:
- Number of frames created can be lower than number registered in labels.csv [See Commit](https://github.com/frgfm/PyroNear/pull/48/commits/258b72d3dc811f0a2bc45804f08457276b5bcc2b)

Isolated, it enables to change extraction strategy independently of the annotation process:
graph_frame_extractor_only_

Here is an example how to use this new object:

```python
    frame_extractor = FrameExtractor(path_to_wildfire,
                                     'jean_lulu_with_seq_01.states.csv',
                                     strategy='random',
                                     n_frames=2)

    labels = (frame_extractor.run(path_to_frames)
                             .get_frame_labels())
```

So far we use this naming convention:
- ` states.csv`: a state is a subsequence of a video composed of frames with constant targets (fire presence, position, confidence and exploitability) : dim=(n_states)
- `labels.csv`: contains extracted frames with the corresponding targets: dim=(n_states x n_frames/state)

Todo list for this PR
- [X] Isolate FrameExtractor (from states to labels+frames)
- [x] Cover with test

Further work for VideoSplitter (to be postponed because the team is currently abundantly annotating)
- Refactor AnnotationParser
- Make AnnotationChecker output checkedStates in addition of checkedLabels

Any feedback is welcome,
Cheers,

---
Current workflow:
graph_original

Suggested workflow:
graph_new_


MateoLostanlen (dismissed):

MateoLostanlen (dismissed):

requirements.txt
```diff
@@ -1,5 +1,9 @@
+opencv-python==4.1.1.26
+pafy==0.5.5
 pandas>=0.25.2
+PyYAML==5.1.2
 torch>=1.2.0
 torchvision>=0.4.0
 tqdm>=4.20.0
 requests>=2.20.0
+youtube-dl==2020.3.24
```
frgfm: I cannot find where `youtube_dl` is required in the PR?
Or perhaps is it a requirement for pafy?

test/fixtures/wildfire_states.csv
```diff
@@ -0,0 +1,15 @@
...
+952_seq1395_1496.mp4,952.mp4,25,True,0,0,1,0,337.429,186.207,0,1395,1496
+952_seq261_310.mp4,952.mp4,25,True,1,0,0,1,1016.803,471.724,0.807,281,310
+952_seq311_410.mp4,952.mp4,25,True,0,0,1,0,559.749,243.762,0,311,410
+952_seq411_463.mp4,952.mp4,25,True,0,0,1,0,444.639,274.232,0,411,463
+952_seq464_757.mp4,952.mp4,25,True,0,0,1,0,396.113,224.577,0,464,757
+952_seq758_778.mp4,952.mp4,25,True,0,0,0,0,312.602,226.834,0,758,778
+952_seq786_799.mp4,952.mp4,25,True,1,0,0,1,1035.987,472.853,0.482,795,799
+952_seq800_815.mp4,952.mp4,25,True,0,0,1,0,975.047,445.768,0,800,815
+952_seq816_1139.mp4,952.mp4,25,True,1,0,0,1,1030.345,479.624,1.573,855,1011
+952_seq816_1139.mp4,952.mp4,25,True,0,0,1,0,1021.317,471.724,7.844,1012,1139
```
frgfm, line 15: Could be put in a gist, rather than on the repo, no?

requirements.txt
```diff
@@ -1,5 +1,9 @@
+opencv-python==4.1.1.26
```
frgfm: I just verified and the functions used from OpenCV are unchanged since version 3. The 4th release is quite a hefty one (noticed when I installed the branch as it was downloading the new release), so I would suggest putting
`opencv-python>=3.4.5.20` instead (which is the most common stable release I've seen from the package).

Then the user will be free to use whichever compatible version from that one on

requirements.txt
```diff
@@ -1,5 +1,9 @@
...
+pafy==0.5.5
```
frgfm: Same here, lower-bound version requirements are preferable here (if `0.5.5` is the minimum compatible version then `>=0.5.5`)

requirements.txt
```diff
@@ -1,5 +1,9 @@
...
 pandas>=0.25.2
+PyYAML==5.1.2
```
frgfm: Same here, lower-bound version requirement would be better :)

test/test_datasets_wildfire_frame_extractor.py
```diff
@@ -0,0 +1,166 @@
...
+# TODO: test when only two frames available and when n_frames > count of available frames
+
+
+class WildFireFrameExtractorTester(unittest.TestCase):
+
+    @staticmethod
+    def download_video_fixtures(path_to_videos):
+        import urllib
+        import yaml
+        import pafy
```
frgfm: As the test will always be executed, better to move imports to the header!

frgfm (changes requested):
Thanks a lot for the PR! Very useful feature, only imports in the test to move to the header, some package requirements to make more flexible, and a fixture file that could be put in a gist.

Thanks again!

requirements.txt
x0s: Yes, that's a requirement for pafy, this is the actual backend that downloads the videos

requirements.txt
x0s: Indeed, OpenCV is a hefty lib, but its size does not seem to change much between version 3 and 4. (Even the version you're suggesting seems heavier):

Screenshot from 2020-04-20 09-38-29

More over, it's always better to have more recent versions as you suggested when reviewing other new dependencies. 
Besides, bug fixes and enhancements of version 3 [have been propagated to version 4](https://github.com/opencv/opencv/wiki/ChangeLog#version3410)
So, I think it would be better to lower bound with the `4.3.0`

test/test_datasets_wildfire_frame_extractor.py
x0s: I put these libs in this well-defined namespace to make it clear which part of the code is going to connect and download videos elsewhere on the net. They may break in the future because of host policy, so we'll have to change.

It's not big deal if you prefer to strictly [follow the PEP 8](https://www.python.org/dev/peps/pep-0008/#imports), I was trying to help the future us ;)

test/fixtures/wildfire_states.csv
x0s: It could be, but I see several cons:
- the data versioning would no longer be ensured (a PR with a new gist breaks history and, old gist can be deleted)
- creates unnecessary coupling (we are not dealing with GigaBytes files but 1.06kB 15 lines CSV still [very readable in browser](https://github.com/frgfm/PyroNear/blob/34a506cf0f701941c7173f859854cd526d01d161/test/fixtures/wildfire_states.csv) and [readability counts](https://www.python.org/dev/peps/pep-0020/)

Keeping this fixture (example) close to the code it uses it, facilitate understanding how to use the feature: it contributes to the documentation.

More over, it won't pollute history because this file is not going to change very often. And if it does, we may want to move it to a _data versioning system_ (a gist is not one).

requirements.txt
frgfm: My bad for the size, I must have mistaken this with a previous version!

However, regarding backward compatibility, I disagree as there might be fixes and improvements in later versions, but regarding the functions being used in this PR, I don't see any behaviour change.

In many projects, backward compatibility is a struggle, but here with the limited dependencies in OpenCV (stable for many versions), I cannot see why forbidding the user to link the dependency to his version, if compatible? If at some point, there is a behaviour difference, we can easily open an issue and just raise the lower bound of version requirement, don't you think?

test/test_datasets_wildfire_frame_extractor.py
frgfm: I would have done the same thing if they were test-only requirements! But since they are also in the installation requirements and that Ctrl+F is quite user-friendly, I saw it as a cleaner option. But you're right, it's far from being critical haha

test/fixtures/wildfire_states.csv
frgfm: I agree with you on data versioning, but with the same reasoning, we would have to push model checkpoint on the repo rather than their URL.
Since we will most likely develop many more features (and thus tests), if we follow the proposed option, we will end up with a fixture folder with more than a dozen of files. I might be wrong but I don't see that as a good option.

For readability, CSV files are all the same readable in browser with a gist ;)

x0s: model checkpoint are not 15lines 1kB readable CSV files. I didn't pretend this reasoning for this little file should be universal. Also your comparison appears in different context. We talk in the context of fixtures in testing not checkpoints when training. Your analogy doesn't apply here.

I think data versioning is the most important here and putting it in a gist is prone to break it.

You say that a fixture folder that may contain a dozen of 1kB CSV files would be a bad option. 
But would a dozen of links pointing to differents gists/account that may break in time is more appealing ?

requirements.txt
x0s: Okay, let's do it

test/test_datasets_wildfire_frame_extractor.py
x0s: Actually, `pafy` and `yaml` are test-only requirements and `urllib` is builtin. Ctrl-F doesn't give you information about namespace dependency. You may have missed my point. I don't want to spend more time on this, so go for it

pyronear/datasets/wildfire/frame_extractor.py
```diff
@@ -71,6 +71,7 @@ def __init__(self,
                              f"Please choose from : {', '.join(self.strategies_allowed)}")
 
         self.states = pd.read_csv(path_to_states)
+        this.states = this.states.astype({"stateStart": int, "stateEnd": int})
```
x0s: As it is documented in the [FrameExtractor docstring](https://github.com/x0s/PyroNear/blob/add-wildfire-frame-extractor/pyronear/datasets/wildfire/frame_extractor.py#L40) and in the tests [with states.csv sample](https://github.com/x0s/PyroNear/blob/add-wildfire-frame-extractor/test/fixtures/wildfire_states.csv) but also [here](https://github.com/x0s/PyroNear/blob/add-wildfire-frame-extractor/test/test_datasets_wildfire_frame_extractor.py#L61), [and here](https://github.com/x0s/PyroNear/blob/add-wildfire-frame-extractor/test/test_datasets_wildfire_frame_extractor.py#L46), [there](https://github.com/x0s/PyroNear/blob/add-wildfire-frame-extractor/test/test_datasets_wildfire_frame_extractor.py#L78) and [there](https://github.com/x0s/PyroNear/blob/add-wildfire-frame-extractor/test/test_datasets_wildfire_frame_extractor.py#L89), it is expected to have ints in the CSV since the states are integer by definition. It is then reasonable to expect that another function that generate such CSV would export the statesStart and stateEnd features as integers not float.

This being said, it makes no arm to enforce this type here, but I would recommend to also cast to integers when exporting this features to csv if that's not the case.

About the implementation, Instead to reading twice these columns to import then to cast type, you directly can do it when reading the CSV this way:
```python
self.states = pd.read_csv(path_to_states, dtype={"stateStart": int, "stateEnd": int})
```

test/fixtures/wildfire_states.csv
frgfm: I agree, but if we are to add more similar tests to the repo, I suggest we find a better solution than just committing csv content files.

x0s (commented):
Thanks for adding a new feature, but I think that's beyond the scope of this PR. That would be better to create a new PR for it. You may be adding it because we are very slow in reviewing then merging it onto master. Anyway, I think we are close to merging it.

I provide you with feedback here:
- Please follow the [PEP8 convention for variables](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)
- Please follow the [Type ints convention](https://www.python.org/dev/peps/pep-0484/) this PR is introducing (to increase stability of the code/ documentation: to answer for instance: what types this function is to expect ?)
- `None` is neither False nor True so if you want it to test if a variable contain None, it is better practice to test it explicitly: `if variable is not None:` /  `if variable is None:`
- Please provide documentation in the dosctrings, this is lacking in `run()` and `_write_frames()`
- If you add a new feature, it is important to cover it with test cases. As you can see, adding this feature [decreased covering of this PR](https://codecov.io/gh/frgfm/PyroNear/pull/48/diff?src=pr&el=tree#D3-218) One may expected to test for possibles values (`None` , differents sizes and some pathological cases: one dimension provided when 2 are expected for instance)
- Please provide a more detailed example illustrating when we need to speed up the code and why resizing would be a good solution.

Answering these could be done in another PRs i guess

requirements.txt
x0s: Done!

requirements.txt
x0s: Done!

pyronear/datasets/wildfire/frame_extractor.py
x0s: Since, I'm the author of this PR, I cannot request changes. 
To prevent the PR to be merged inadvertently without taking into account my request changes, I converted the PR to draft.

That's one more reason to avoid adding a new feature in a PR in review state. It's not big deal, though, Paradoxically, I will mark it again as ready for review once the "author"'s review would be done ^^

frgfm: @MateoLostanlen would you mind taking a look at it please? I'll request changes on behalf of @x0s

frgfm (changes requested):
@MateoLostanlen could you take a look at @x0s comment on the `frame_extractor` please?

pyronear/datasets/wildfire/frame_extractor.py
MateoLostanlen: Hi ! This is a typing error but I have already correct it in the following commit :
https://github.com/frgfm/PyroNear/pull/48/commits/2f8fd23bce207f977162f33438cd10cff33950f2

x0s: @MateoLostanlen : My review was not about the typing error, maybe read it again ;) 
https://github.com/frgfm/PyroNear/pull/48#discussion_r415638269

MateoLostanlen: Okay, sorry, I read that a little fast.

MateoLostanlen (dismissed):

pyronear/datasets/wildfire/frame_extractor.py
frgfm: Perhaps @x0s you might want to add commit to revert the changes of @MateoLostanlen on this section so we can merge the PR?
Or is there anything else blocking?

x0s: @frgfm: As I pointed on slack, since my motherboard became defective, I have limited access to the repository.
Nevertheless, last week, I recommended  to @MateoLostanlen to revert its commits in favor of a new PR if needed. For instance:
```
git revert --no-commit b2fefbc^..34a506c
git commit -m "Revert reviewer commits in favor of another PR"
```
What else is blocking ?

frgfm: I'm good with that, it was just to avoid a multi-author PR situation. But yes, @MateoLostanlen could you revert this change please?

x0s: @frgfm : I don't get what you mean by "multi-author PR situation", can you elaborate ?

frgfm (approved):
All good!

PR #50, opened 2020-04-13

MateoLostanlen:


frgfm:
@MateoLostanlen there seems to be some conflicts on your branch, we won't be able to merge. Could you either pull master and open a PR from a clean branch, or merge master into yours perhaps? That should fix the issue!

x0s:
@MateoLostanlen: Thanks for the PR. Can you provide details about how you get the error ? (code, commands and context so we **can reproduce it**)

As @frgfm suggested, it could be great to rebase/merge your branch on top of `master` to see the new changes

PR #53, opened 2020-04-15

MateoLostanlen:


x0s:
Following our discussions and several attempts, the problem does not not seem to come from github UI.
Maybe, the problem comes from how github refresh the diff; [as suggested here](https://stackoverflow.com/a/46782679/3581903), please try to:
- Click on **Edit** on top-right corner
- Change **base** branch to your **fork's master** branch, save and refresh
- Click on **Edit** again and switch back the base branch to the branch your initially want to merge: **type-correction**

Refresh and let's see what happens!

MateoLostanlen:
Hello, I tried your solution unfortunatly I can't choose a branch on my fork as a base

image

x0s:
It seems like the SHA hashes of the old commits not expected in the current diff have been changed between your branch and master. We can see that your github account has been added as a **committer** to all these commits informations. I don't how its happens. You may have changed something in your local git config that mingled with pulling process.

Let's see where it diverged:
- see on master:
git_mateo_1
- see on your branches (both `master` and `type-correction`:  
git_mateo_2

You can see the hash and committers names have changed.

Since your `master` has this divergence, every other PR would behave the same way (trying to commit these "non-changes"

Several possibilities here:
- You can reset hard your master branch on the "last correct commit" then rebase on top of main repo master branch (with the correct authors and commiters).
- backup your master branch, and create another one from main repo master. check the history and create again a new branch type-correction` from your fresh up-to-date master, where you'll be able to add your commit

PR #55, opened 2020-04-19
style: Added encoding in headers

frgfm:
This PR aims at:
- adding utf8 encoding when it was missing in header to avoid issues
- switched to Python 3 class definitions style
- switched from absolute to relative import when it was not needed (absolute could use existing package installation even if they do not share the same codebase)

### DIFF
--- a/pyronear/datasets/wildfire/fire_labeler.py
+++ b/pyronear/datasets/wildfire/fire_labeler.py
@@ -1,3 +1,6 @@
+#!usr/bin/python
+# -*- coding: utf-8 -*-
+
 import re
 from itertools import combinations
 
--- a/pyronear/datasets/wildfire/fire_labeler.py
+++ b/pyronear/datasets/wildfire/fire_labeler.py
@@ -7,7 +7,7 @@ from itertools import combinations
 import numpy as np
 
 
-class FireLabeler(object):
+class FireLabeler:
     """Automatically labelize WildFire dataset based on video descriptions
 
     It will create a new column containing Fire ids that try to identify the videos
--- a/pyronear/datasets/wildfire/fire_labeler.py
+++ b/pyronear/datasets/wildfire/fire_labeler.py
@@ -7,7 +7,7 @@ from itertools import combinations
 import numpy as np
 
 
-class FireLabeler:
+class FireLabeler(object):
     """Automatically labelize WildFire dataset based on video descriptions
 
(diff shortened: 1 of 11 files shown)


pyronear/datasets/wildfire/fire_labeler.py
```diff
@@ -1,10 +1,13 @@
...
+# -*- coding: utf-8 -*-
+
 import re
 from itertools import combinations
 
 import numpy as np
 
 
-class FireLabeler:
+class FireLabeler(object):
```
x0s: All classes are new-style in python 3, so we no longer need to inherit from object.

pyronear/datasets/wildfire/wildfire.py
```diff
@@ -72,7 +75,7 @@ def _get_targets(self, index):
         return torch.from_numpy(self.metadata[self.target_names].iloc[index].values)
 
 
-class WildFireSplitter:
+class WildFireSplitter(object):
```
x0s: Same all classes are new-style in python 3, so we no longer need to inherit from object.

pyronear/datasets/wildfire/fire_labeler.py
```diff
@@ -1,10 +1,13 @@
+#!usr/bin/python
```
x0s: Why do you add the shebang ? this file is not intended to be used as a script

x0s: same remark for all new shebang in the current PR

x0s (changes requested):
Thanks for the PR. Quickly:
- Class definition are already new-style
- The files concerned in this PR are not intended to be used as scripts
- Okay for relative imports

pyronear/datasets/wildfire/fire_labeler.py
frgfm: My bad, you are very correct

pyronear/datasets/wildfire/fire_labeler.py
frgfm: Oops, always copy pasted it with the encoding, you are correct!

frgfm:
> Thanks for the PR. Quickly:
> 
> * Class definition are already new-style
> * The files concerned in this PR are not intended to be used as scripts
> * Okay for relative imports

Thanks for the review, just made the suggested modifications!

x0s (approved):

PR #56, opened 2020-04-21

MateoLostanlen:


x0s:
To understand the problem you're trying to solve, can you give us a snippet and context to reproduce the error ?

MateoLostanlen:
I'm closing this PR for now, I'm not able to reproduce the error

PR #57, opened 2020-04-30
fix: Fixed version attribute import

### LINKED ISSUE #49: What is "from .version import __version__" in __init__.py?
## 🐛 Bug

The import "from .version import __version__" in __init__.py? generate the following error :

ModuleNotFoundError: No module named 'pyronear.version'

I use to comment it, it seems to make no difference. Can we remove it ?

frgfm:
This PR aims at:
- fixing the version attribute access in dev mode

Trying to access `pyronear.__version__` will throw an AttributeError if the package is not installed, while bare imports of top-level module `import pyronear` or `from pyronear import *` will not throw any issue.


pyronear/__init__.py
```diff
@@ -1,4 +1,8 @@
-from .version import __version__
 from pyronear import datasets
 from pyronear import models
 from pyronear import utils
+
+try:
+    from .version import __version__  # noqa: F401
+except ImportError:
+    pass
```
x0s: When you catch an exception, you have to treat it. If there is no way to treat it here, then it's better to not have it happen (if possible).
And that seems possible: Following the Official guidelines (Third point for instance, as suggested [here](https://github.com/frgfm/PyroNear/issues/49#issuecomment-607196633) in #49, will prevent this ImportError to happen. So why not following these guidelines ?

x0s (changes requested):
Thanks for treating #49. Why not simply follow official guidelines ?

pyronear/__init__.py
frgfm: Hesitated to do this! The only inconvenient I saw was that we would have two version indicators in the repo (in `version.py` and `setup.py`)

But I agree the error catching is not among the cleanest!

x0s: I don't understand why our library should differ with major ones like pip with regards to version number ? for instance
- writing version in init (no import): https://github.com/pypa/pip/blob/master/src/pip/__init__.py
- reading in `setup.py` from that init: https://github.com/pypa/pip/blob/master/setup.py

frgfm: Happy to change it, I didn't see any issue in this as this choice is also made by many other "major" ones such as [torchvision](https://github.com/pytorch/vision/blob/master/torchvision/__init__.py#L14) or [pytorch](https://github.com/pytorch/pytorch/blob/master/torch/__init__.py#L23).

x0s: You're right, but since there are lot of ways to avoid silencing an ImportError Exception, I don't think pytorch made a good decision on this.

Thanks, then

x0s (approved):

PR #60, opened 2020-04-30
fix: Fixed images format to tensor by default.

### LINKED ISSUE #58: Wrong batch images format when iterating through DataLoader
## 🐛 Bug

https://github.com/frgfm/PyroNear/blob/61d5d527187a6f6b0696da0bc199c1a80a2ee2c0/pyronear/datasets/utils.py#L19

Within VisionMixin class, 'load_image' method's 'to_tensor' parameter should be set to 'True' by default.  

## To Reproduce

1. Iterate through dataLoader 
```
for epoch in range(num_epochs):
    for i, (images, labels) in enumerate(wildfire_loader_train):
```
which produces : 

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-12-4d31344fa390> in <module>
      4 iter = 0
      5 for epoch in range(num_epochs):
----> 6     for i, (images, labels) in enumerate(wildfire_loader_train):
      7 
      8         labels = labels.squeeze()

/usr/local/lib/python3.7/site-packages/torch/utils/data/dataloader.py in __next__(self)
    344     def __next__(self):
    345         index = self._next_index()  # may raise StopIteration
--> 346         data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
    347         if self._pin_memory:
    348             data = _utils.pin_memory.pin_memory(data)

/usr/local/lib/python3.7/site-packages/torch/utils/data/_utils/fetch.py in fetch(self, possibly_batched_index)
     45         else:
     46             data = self.dataset[possibly_batched_index]
---> 47         return self.collate_fn(data)

/usr/local/lib/python3.7/site-packages/torch/utils/data/_utils/collate.py in default_collate(batch)
     77     elif isinstance(elem, container_abcs.Sequence):
     78         transposed = zip(*batch)
---> 79         return [default_collate(samples) for samples in transposed]
     80 
     81     raise TypeError(default_collate_err_msg_format.format(elem_type))

/usr/local/lib/python3.7/site-packages/torch/utils/data/_utils/collate.py in <listcomp>(.0)
     77     elif isinstance(elem, container_abcs.Sequence):
     78         transposed = zip(*batch)
---> 79         return [default_collate(samples) for samples in transposed]
     80 
     81     raise TypeError(default_collate_err_msg_format.format(elem_type))

/usr/local/lib/python3.7/site-packages/torch/utils/data/_utils/collate.py in default_collate(batch)
     79         return [default_collate(samples) for samples in transposed]
     80 
---> 81     raise TypeError(default_collate_err_msg_format.format(elem_type))

TypeError: default_collate: batch must contain tensors, numpy arrays, numbers, dicts or lists; found <class 'PIL.Image.Image'>
```

## Expected behavior

If set to True, image is transformed into a pytorch tensor as expected.

Akilditu:
This PR aims at loading images from a path into a torch tensor.

Fixes #58

### DIFF
--- a/pyronear/datasets/utils.py
+++ b/pyronear/datasets/utils.py
@@ -16,7 +16,7 @@ from torchvision.datasets.utils import check_integrity
 class VisionMixin:
     """Class to provide re-usabled functions to classes dealing this Vision (ie: VisionDataset)"""
     @staticmethod
-    def load_image(path, to_tensor=False):
+    def load_image(path, to_tensor=True):
         """Load an image from a path into a (PIL Image) or a (pytorch Tensor)"""
         # open path as file to avoid ResourceWarning (https://github.com/python-pillow/Pillow/issues/835)
         with open(path, 'rb') as f:


MateoLostanlen:
I don't think it's a good idea to do that because we're going to have a problem applying the other transformations. Indeed torchvision.transforms takes a pillow image as input and not a tensor.

https://pytorch.org/docs/stable/torchvision/transforms.html

MateoLostanlen:
Maybe we can set the default transformation to to_tensor instead of None.

x0s (changes requested):
Thanks for the PR!

It's always tricky to change default value because one may break already existing code.

But, here, it depends what we want.
1. For now, one can load this Dataset of Image and when iterating, one gets Images. This is intuitive.
2. If one wants to transform the images with pytorch (therefore [converting it to tensor like it is documented here](https://github.com/frgfm/PyroNear/blob/master/test/test_datasets_wildfire_split.py#L61), one gets transformed images as expected.

If we enforce `to_tensor` to ` True` in WildFireDataset, we loose 1. (capability of displaying genuine images)
If we enforce `transforms` to ToTensor, we need to enforce `None` to get Images, it would now seems counter-intuitive. 

If we let it as it is, the `load_images` feature _loads images_ by default but it capable of converting it in Tensor. More over, if we want to transform images, we simply should append ToTensor to the sequence of transforms like [it is advised in Official tutorial](https://pytorch.org/tutorials/beginner/data_loading_tutorial.html#iterating-through-the-dataset)
One more advantage of keeping as it is, one can try different sets of transforms on images before settling, [like in the tutorial](https://pytorch.org/tutorials/beginner/data_loading_tutorial.html#compose-transforms), it's quite handy.

Thanks @Akilditu for raising this discussion, it will help us to get in tune with the training workflow ;)

What's your thoughts on this ?


Besides, You can launch the tests locally to see if already existing behaviour may break:
`python -m unittest discover -vs test/`

Akilditu:
Indeed @MateoLostanlen  and @x0s thanks for your remarks on this. 

Applying `transforms.ToTensor()` while creating `WildFireDataset` using the `transform` parameter  seems to be the better option [here](https://github.com/frgfm/PyroNear/blob/acd9cc8b513abcadec80727d85a1a75358e8be44/pyronear/datasets/wildfire/wildfire.py#L34-L35).

But now is `to_tensor` still relevant [here](https://github.com/frgfm/PyroNear/blob/acd9cc8b513abcadec80727d85a1a75358e8be44/pyronear/datasets/utils.py#L19) ?

MateoLostanlen:
@Akilditu  We can leave this option. Can we close this PR ?

Akilditu:
@MateoLostanlen, yep closing PR right now !

PR #63, opened 2020-05-01
Allow test ratio to zero in WildFireSplitter

### LINKED ISSUE #61: [DataSet] Allow test ratio to zero in WildFireSplitter
## Motivation

At the moment the input ratios of WildFireSplitter cannot be equal to zero. I would like to add this option. Indeed we are going to do a lot of tests with different extraction methods (number of frames, spacing between frames ...). As the frame extraction is long I think we will save a lot of time if we don't have to do it every time for the test set. 

I will try to propose a PR to add this option today

MateoLostanlen:
This PR aims to answer the issue : [DataSet] Allow test ratio to zero in WildFireSplitter #61

As explain in the issue :

At the moment the input ratios of WildFireSplitter cannot be equal to zero. I would like to add this option. Indeed we are going to do a lot of tests with different extraction methods (number of frames, spacing between frames ...). As the frame extraction is long I think we will save a lot of time if we don't have to do it every time for the test set.

PS: I have also commented the warinig

```
"Labeling. Maybe try to label the fire again
Distinct values of ids(340 ≠ 420)
f" ≠ {wildfire.metadata['fire_id'].max() + 1})", Warning)"
```

We haven't annotated the entire Dataset so we're going to get this warning every time.

### DIFF
--- a/pyronear/datasets/wildfire/split_strategy.py
+++ b/pyronear/datasets/wildfire/split_strategy.py
@@ -72,7 +72,8 @@ class ExhaustSplitStrategy(SplitStrategy):
         if ratios['test'] > 0:
             fire_ids = {'train': self._get_fire_ids_for_one_split(n_samples_train),
                         'val': self._get_fire_ids_for_one_split(n_samples_val)}
-            fire_ids['test'] = [id_ for id_ in self._fire_id_to_size if id_ not in (fire_ids['train'] + fire_ids['val'])]
+            fire_ids['test'] = [id_ for id_ in self._fire_id_to_size
+                                if id_ not in (fire_ids['train'] + fire_ids['val'])]
             # Finish exhaustion
             for fire_id_test in fire_ids['test']:
                 del self._fire_id_to_size_to_exhaust[fire_id_test]
@@ -83,7 +84,7 @@ class ExhaustSplitStrategy(SplitStrategy):
             # Finish exhaustion
             for fire_id_test in fire_ids['val']:
                 del self._fire_id_to_size_to_exhaust[fire_id_test]
-            fire_ids['test']=[]
+            fire_ids['test'] = []
 
         n_samples_remaining = len(self._fire_id_to_size_to_exhaust)
         if n_samples_remaining != 0:
--- a/pyronear/datasets/wildfire/split_strategy.py
+++ b/pyronear/datasets/wildfire/split_strategy.py
@@ -69,12 +69,21 @@ class ExhaustSplitStrategy(SplitStrategy):
         self._fire_id_to_size_to_exhaust = copy.deepcopy(self._fire_id_to_size)
 
         # Let's get
-        fire_ids = {'train': self._get_fire_ids_for_one_split(n_samples_train),
-                    'val': self._get_fire_ids_for_one_split(n_samples_val)}
-        fire_ids['test'] = [id_ for id_ in self._fire_id_to_size if id_ not in (fire_ids['train'] + fire_ids['val'])]
-        # Finish exhaustion
-        for fire_id_test in fire_ids['test']:
-            del self._fire_id_to_size_to_exhaust[fire_id_test]
+        if ratios['test'] > 0:
+            fire_ids = {'train': self._get_fire_ids_for_one_split(n_samples_train),
+                        'val': self._get_fire_ids_for_one_split(n_samples_val)}
+            fire_ids['test'] = [id_ for id_ in self._fire_id_to_size if id_ not in (fire_ids['train'] + fire_ids['val'])]
+            # Finish exhaustion
+            for fire_id_test in fire_ids['test']:
+                del self._fire_id_to_size_to_exhaust[fire_id_test]
+
+        else:
+            fire_ids = {'train': self._get_fire_ids_for_one_split(n_samples_train)}
+            fire_ids['val'] = [id_ for id_ in self._fire_id_to_size if id_ not in fire_ids['train']]
+            # Finish exhaustion
+            for fire_id_test in fire_ids['val']:
+                del self._fire_id_to_size_to_exhaust[fire_id_test]
+            fire_ids['test']=[]
 
         n_samples_remaining = len(self._fire_id_to_size_to_exhaust)
         if n_samples_remaining != 0:
--- a/pyronear/datasets/wildfire/wildfire.py
+++ b/pyronear/datasets/wildfire/wildfire.py
@@ -158,10 +158,10 @@ class WildFireSplitter:
         """
         self.wildfire = wildfire
         # Some checks first
-        if wildfire.metadata['fire_id'].nunique() != wildfire.metadata['fire_id'].max() + 1:
-            warnings.warn(f"Inconsistent Fire Labeling. Maybe try to label the fire again\n"
-                          f"Distinct values of ids({wildfire.metadata['fire_id'].nunique()}"
-                          f" ≠ {wildfire.metadata['fire_id'].max() + 1})", Warning)
+        # if wildfire.metadata['fire_id'].nunique() != wildfire.metadata['fire_id'].max() + 1:
+        #     warnings.warn(f"Inconsistent Fire Labeling. Maybe try to label the fire again\n"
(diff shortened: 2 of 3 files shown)


frgfm:
Hi @MateoLostanlen,

Thanks for the PR! Is this ready for review?
If so, add some reviewers to it or we don't get notified :)

MateoLostanlen:
Yes it's ready :)

test/test_datasets_wildfire_split.py
```diff
@@ -93,6 +93,15 @@ def test_inconsistent_ratios_raise_exception(self):
...
             WildFireSplitter(ratios)
 
+    def test_splitting_with_test_to_zero(self):
+        ratios = {'train': 0.81, 'val': 0.19, 'test': 0}
+
+        splitter = WildFireSplitter(ratios, seed=42)
+        splitter.fit(self.wildfire)
+
+        for (set_, ratio_) in splitter.ratios_.items():
+            self.assertAlmostEqual(ratio_, ratios[set_], places=2)
```
frgfm, line 103: Is it required to have so few decimals in `assertAlmostEqual`?
If the precision can be determined using the size of the dataset to be more accurate, that would be even better!

pyronear/datasets/wildfire/wildfire.py
```diff
@@ -158,10 +158,10 @@ def fit(self, wildfire):
...
         self.wildfire = wildfire
         # Some checks first
-        if wildfire.metadata['fire_id'].nunique() != wildfire.metadata['fire_id'].max() + 1:
-            warnings.warn(f"Inconsistent Fire Labeling. Maybe try to label the fire again\n"
-                          f"Distinct values of ids({wildfire.metadata['fire_id'].nunique()}"
-                          f" ≠ {wildfire.metadata['fire_id'].max() + 1})", Warning)
+        # if wildfire.metadata['fire_id'].nunique() != wildfire.metadata['fire_id'].max() + 1:
+        #     warnings.warn(f"Inconsistent Fire Labeling. Maybe try to label the fire again\n"
+        #                   f"Distinct values of ids({wildfire.metadata['fire_id'].nunique()}"
+        #                   f" ≠ {wildfire.metadata['fire_id'].max() + 1})", Warning)
```
frgfm: If commented, do we need to keep this section at all?

pyronear/datasets/wildfire/split_strategy.py
```diff
@@ -69,12 +69,22 @@ def split(self, dataset, ratios, seed=42):
...
+            for fire_id_test in fire_ids['test']:
+                del self._fire_id_to_size_to_exhaust[fire_id_test]
+
+        else:
+            fire_ids = {'train': self._get_fire_ids_for_one_split(n_samples_train)}
+            fire_ids['val'] = [id_ for id_ in self._fire_id_to_size if id_ not in fire_ids['train']]
+            # Finish exhaustion
+            for fire_id_test in fire_ids['val']:
+                del self._fire_id_to_size_to_exhaust[fire_id_test]
+            fire_ids['test'] = []
```
frgfm, line 87: Could you write in #61 the command you used to produce the error this is trying to solve?

So that we can test it :)

frgfm (commented):
Thanks a lot! Would you mind providing us in #61 with the exact commands you used to produce the error this PR is addressing please?

The commented section might removed, and I wrote a small question about the test precision. Let me know what you think!

pyronear/datasets/wildfire/wildfire.py
x0s: This warning is here for some reason. Splitting this dataset suppose firelabeling was correctly made.
If it wasn't, a warning is raised to warn the user. It could be difficult to see it from outside the function.

Besides, If you want to silence warnings when using this code in your script/notebook in case too much information displayed bother you, you can do it this way: (no need to silence warnings for everyone)
```python
import warnings
warnings.filterwarnings("ignore")
```

test/test_datasets_wildfire_split.py
x0s: These are estimated ratios and 2 decimals is enough here because most important constraint is to group frames referring to same fire.  One may improve assigning algorithm, but biasing in order to be better at respecting ratios size is not a good idea IMO. This is not the kind of biases that would improve predictability.

pyronear/datasets/wildfire/wildfire.py
MateoLostanlen: For me a warning that is raised under any circumstances doesn't really make sense. We can eventually put it back on if we annotate the entire dataset. We can also modify it by considering what was actually annotated

pyronear/datasets/wildfire/split_strategy.py
MateoLostanlen: Done :)

pyronear/datasets/wildfire/wildfire.py
x0s: Indeed, It does raise a warning under any circumstances corresponding to the same misuse of metadata.

If you instantiate a `WildFireDataset` object with a metadata file, it should cover exactly the values of this dataset, not more not less. Here a warning is raised because such inconsistency was detected.It is important, because this a a responsibility for the splitter not to mess the fire ids. 

Also, the count of frames in the directory used to init WildFireDataset does not prevail on the metadata to give information on the dataset (actually, this is the role of _metadata_ to do so).
This is even more dangerous because using a metadata not reflecting the data, give [wrong length of WildFireDataset !](https://github.com/frgfm/PyroNear/blob/master/pyronear/datasets/wildfire/wildfire.py#L52) so we can expect that some splitting strategies may process the dataset in unexpected ways !
Following this conversation, I think precising this in the documentation and testing metadata consistency over the data it covered in WildFireDataset may help us.

So, If you work on a subset of Wildfire today, it may be better working with a subset of the metadata.csv, and it should work well. At some point you should have made this distinction.

MateoLostanlen: Okay, I'm sorry, I think I don't understand something. I'm gonna explain my thinking to you to see if I'm missing something. I added the FireLabels to my dataset, the fires are labeled from the names of the 1020 videos, which gives 420 different fires, which is what we find in the warning:

`wildfire.metadata['fire_id'].max()`

But since we didn't annotate all the dataset the number of fireid actually present in the annotations is 340.  `wildfire.metadata['fire_id'].nunique()`
What raised the warning:

`Distinct values of ids(340 ≠ 420)`

Am I using the fireLabeler in the wrong way? Should I have deleted the unlabeled videos from my video list?

x0s: Thank you for detailing your workflow, I understand now more clearly what you're trying to achieve.

I recommend to only work with the data available. Checks should pass and it would be more convenient for you to work. Here is a suggestion.

Let's see the flow [from the videos to the WildFireDataset](https://user-images.githubusercontent.com/6331909/77332236-182adf80-6d22-11ea-897d-6449a02a77fa.png)

At beginning, we can have:
- all videos in a directory
- checkedstates **covering only annotated data**
- metadata _covering all videos_ <-- this needs to be changed to **cover only annotated data**

So if you, create a new metadata.csv from the original(exhaustive) one, you tell the other objects (WildFireDataset, FireLabeler, WildFiresplitter...) the videos you are going to use.
For instance, [using a merge](https://pandas.pydata.org/pandas-docs/stable/user_guide/merging.html#database-style-dataframe-or-named-series-joining-merging) between checkstates and original metadata.

The role of metadata is to _provide information of the data (videos/frames) you need to process_. The other objects will process data _base on these information_, so that's important metadata only contain what's needed/available.
Concerning the videos/frames, you shouldn't need to subselect what's needed/delete unannotated videos because the objects(splitter...) will be able to filter based on metadata.csv.

Of course, we can improve this process. It is very valuable to have your feedback, tell me if this helps.
Thanks,

MateoLostanlen: Okay, I'll work on my workflow. I'll put the warning back on. Thanks

frgfm (approved):
Thanks for the PR! Looks good to me!

PR #64, opened 2020-05-02
Train WildFire

MateoLostanlen:
The purpose of this PR is to add a basic training loop for the wildfire dataset. This is far from being optimal in terms of performance but it can be considered as a first baseline.


pyronear/datasets/wildfire/split_strategy.py
```diff
@@ -69,12 +69,22 @@ def split(self, dataset, ratios, seed=42):
...
+            for fire_id_test in fire_ids['test']:
+                del self._fire_id_to_size_to_exhaust[fire_id_test]
+
+        else:
+            fire_ids = {'train': self._get_fire_ids_for_one_split(n_samples_train)}
+            fire_ids['val'] = [id_ for id_ in self._fire_id_to_size if id_ not in fire_ids['train']]
+            # Finish exhaustion
+            for fire_id_test in fire_ids['val']:
+                del self._fire_id_to_size_to_exhaust[fire_id_test]
+            fire_ids['test'] = []
```
frgfm: Isn't this supposed to be in #63 ?

pyronear/datasets/wildfire/wildfire.py
```diff
@@ -158,10 +158,10 @@ def fit(self, wildfire):
...
         self.wildfire = wildfire
         # Some checks first
-        if wildfire.metadata['fire_id'].nunique() != wildfire.metadata['fire_id'].max() + 1:
-            warnings.warn(f"Inconsistent Fire Labeling. Maybe try to label the fire again\n"
-                          f"Distinct values of ids({wildfire.metadata['fire_id'].nunique()}"
-                          f" ≠ {wildfire.metadata['fire_id'].max() + 1})", Warning)
+        # if wildfire.metadata['fire_id'].nunique() != wildfire.metadata['fire_id'].max() + 1:
+        #     warnings.warn(f"Inconsistent Fire Labeling. Maybe try to label the fire again\n"
+        #                   f"Distinct values of ids({wildfire.metadata['fire_id'].nunique()}"
+        #                   f" ≠ {wildfire.metadata['fire_id'].max() + 1})", Warning)
```
frgfm: Same here

test/test_datasets_wildfire_split.py
```diff
@@ -93,6 +93,15 @@ def test_inconsistent_ratios_raise_exception(self):
...
 
+    def test_splitting_with_test_to_zero(self):
+        ratios = {'train': 0.81, 'val': 0.19, 'test': 0}
+
+        splitter = WildFireSplitter(ratios, seed=42)
+        splitter.fit(self.wildfire)
+
+        for (set_, ratio_) in splitter.ratios_.items():
+            self.assertAlmostEqual(ratio_, ratios[set_], places=2)
+
```
frgfm: Same here

frgfm (changes requested):
Thanks for the PR!
There seems to be an issues with overlapping modifications with #63. Could you take a look please?

pyronear/datasets/wildfire/split_strategy.py
MateoLostanlen: Yes it's in #63, I wrote this PR on top of #63. So we need to merge #63 before this one

references/classification/WildFire/torch/train.py
```diff
@@ -0,0 +1,311 @@
...
+            args.device = 'cuda:0'
+        else:
+            args.device = 'cpu'
+
+    #Create Dataloaders
+
+    normalize = transforms.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225])
+
+    train_transforms = transforms.Compose([
+        transforms.RandomResizedCrop(size=args.resize // 7, scale=(0.8, 1.0)),
```
frgfm: Sounds like a pretty big loss of information with a relative output size of  `1/7`. Perhaps we should directly resize to target size?

references/classification/WildFire/torch/train.py
```diff
@@ -0,0 +1,311 @@
...
+        transforms.RandomRotation(degrees=15),
+        transforms.ColorJitter(),
+        transforms.RandomHorizontalFlip(),
+        transforms.CenterCrop(size=args.resize),  # Image net standards
```
frgfm: Generally CenterCrop is used for eval but not train set. Having two crop transformations seems unnecessary. It might be better to stick with RandomResizedCrop but discard CenterCrop

references/classification/WildFire/torch/train.py
```diff
@@ -0,0 +1,311 @@
...
+        transforms.ToTensor(),
+        normalize
+    ])
+
+    val_transforms = transforms.Compose([
+        transforms.Resize(size=args.resize // 7),
```
frgfm: same here for the small resize

frgfm (changes requested):
Thanks for the PR! A few changes and some transformation choices that I'm having doubts about, let me know your thoughts about the comments!

references/classification/WildFire/torch/train.py
MateoLostanlen: It was a mistake, I didn't want to lose so much information. Thank you, that's corrected. I also deleted one of the two crops.

references/classification/WildFire/torch/train.py
```diff
@@ -158,18 +158,16 @@ def main(args):
...
         transforms.RandomHorizontalFlip(),
-        transforms.CenterCrop(size=args.resize),  # Image net standards
         transforms.ToTensor(),
         normalize
     ])
 
     val_transforms = transforms.Compose([
-        transforms.Resize(size=args.resize // 7),
-        transforms.CenterCrop(size=args.resize),
+        transforms.Resize(size=args.resize),
```
frgfm, line 170: Small difference here: `RandomResizedCrop` does crop **then** resize the rop.
But here we need the CenterCrop after the Resize. The reason behind it (https://pytorch.org/docs/stable/torchvision/transforms.html#torchvision.transforms.Resize) : if you pass an int to Resize, it keeps the aspect ratio and resize the smallest side to `args.resize`. If you pass a tuple it will resize the entire image to target shape (potentially changing the aspect ratio).

On eval, you probably don't want to stretch the image, but you need them to be all of the same size to batch them. So keep the `transforms.Resize` as is but add back the CenterCrop as well to make sure the image is not stretched !

frgfm (commented):
Small change to make for eval transformations but the rest looks good!

references/classification/WildFire/torch/train.py
MateoLostanlen: I should resize using a tuple to keep the ratio and then do a centercrop?

frgfm: I was rather suggesting doing resize with an int (`args.resize`) then doing a `CenterCrop(arg.resize)`.
If the image is squared, CenterCrop won't do a thing, if it isn't, it will make it square by cropping

MateoLostanlen: oh yes sorry, Done !

pyronear/datasets/wildfire/split_strategy.py
frgfm: Could you cancel out this modif please?

test/test_datasets_wildfire_split.py
frgfm: Could you cancel out this modification? The rest of the PR looks good!

frgfm (commented):
Only some side modifications to cancel out, the rest looks good!

Akilditu (commented):
Thanks for the PR! 

Added some comments to let the ability of optimizer change and tensor shape for targets.

references/classification/WildFire/torch/train.py
```diff
@@ -0,0 +1,310 @@
...
+        model.load_state_dict(torch.load(args.resume)['model'])
+
+    # Send to device
+    model.to(args.device)
+
+    # Loss function
+    criterion = nn.CrossEntropyLoss()
+
+    # optimizer
+    optimizer = optim.Adam(model.parameters(),
```
Akilditu, line 213: I would recommend to let the ability to use other Optimizers than Adam

references/classification/WildFire/torch/train.py
```diff
@@ -0,0 +1,310 @@
...
+    os.environ['PYTHONHASHSEED'] = str(seed)
+    np.random.seed(seed)
+    torch.manual_seed(seed)
+    torch.cuda.manual_seed(seed)
+    torch.cuda.manual_seed_all(seed)  # if you are using multi-GPU.
+    torch.backends.cudnn.benchmark = False
+    torch.backends.cudnn.deterministic = True
+
+
+def train_batch(model, x, target, optimizer, criterion):
```
Akilditu, line 44: Does target needs to be squeeze() here ?

frgfm: Not sure why it would? The model will need the batch dimension, so we only `squeeze` it if it was already `unsqueeze`. If I'm not mistaking, it's not needed then

references/classification/WildFire/torch/train.py
frgfm: Agreed, but perhaps we can introduce this in another PR?

MateoLostanlen: Yes let's do that

references/classification/WildFire/torch/train.py
MateoLostanlen: It depend of your dataloader, but in this script there is no need to squeeze the target

pyronear/datasets/wildfire/split_strategy.py
MateoLostanlen: Done !

frgfm (approved):
Thanks a lot for the changes!

PR #66, opened 2020-05-17
Add SubSsampler dataset

MateoLostanlen:
Cette PR introduit mon wildfireSubSampler DataSet, le but de ce dataset est de combiner K frame d’une même séquence. Ce Dataset permet d’entrainer une version modifiée du ResNet18, le ssResNet18. Ce model passe les K frame dans la partie convolution du model indépendamment puis les assemble dans la partie fully connected pour prédire s’il y a du feu ou non. L’idée et d’ajouter une sorte d’information temporelle pour améliorer les prédictions


frgfm:
I'll wait up on #63 to be merged to ensure the tests are running correctly! If I may, for next PRs, try to make sure that they are not inter-dependent, that will make things more straightforward for reviewing :+1:

MateoLostanlen:
Ok, sorry for that !

pyronear/datasets/wildfire/split_strategy.py
```diff
@@ -69,12 +69,22 @@ def split(self, dataset, ratios, seed=42):
...
+            for fire_id_test in fire_ids['test']:
+                del self._fire_id_to_size_to_exhaust[fire_id_test]
+
+        else:
+            fire_ids = {'train': self._get_fire_ids_for_one_split(n_samples_train)}
+            fire_ids['val'] = [id_ for id_ in self._fire_id_to_size if id_ not in fire_ids['train']]
+            # Finish exhaustion
+            for fire_id_test in fire_ids['val']:
+                del self._fire_id_to_size_to_exhaust[fire_id_test]
+            fire_ids['test'] = []
```
frgfm: You might want to cancel the modification here since the related PR was merged

pyronear/datasets/wildfire/wildfire.py
```diff
@@ -158,10 +158,10 @@ def fit(self, wildfire):
...
         self.wildfire = wildfire
         # Some checks first
-        if wildfire.metadata['fire_id'].nunique() != wildfire.metadata['fire_id'].max() + 1:
-            warnings.warn(f"Inconsistent Fire Labeling. Maybe try to label the fire again\n"
-                          f"Distinct values of ids({wildfire.metadata['fire_id'].nunique()}"
-                          f" ≠ {wildfire.metadata['fire_id'].max() + 1})", Warning)
+        # if wildfire.metadata['fire_id'].nunique() != wildfire.metadata['fire_id'].max() + 1:
+        #     warnings.warn(f"Inconsistent Fire Labeling. Maybe try to label the fire again\n"
+        #                   f"Distinct values of ids({wildfire.metadata['fire_id'].nunique()}"
+        #                   f" ≠ {wildfire.metadata['fire_id'].max() + 1})", Warning)
```
frgfm: Same here

test/test_datasets_wildfire_split.py
```diff
@@ -93,6 +93,15 @@ def test_inconsistent_ratios_raise_exception(self):
...
 
+    def test_splitting_with_test_to_zero(self):
+        ratios = {'train': 0.81, 'val': 0.19, 'test': 0}
+
+        splitter = WildFireSplitter(ratios, seed=42)
+        splitter.fit(self.wildfire)
+
+        for (set_, ratio_) in splitter.ratios_.items():
+            self.assertAlmostEqual(ratio_, ratios[set_], places=2)
+
```
frgfm: same here

frgfm (changes requested):
Thanks again! Would you mind opening a different PR for the model?

pyronear/models/ssresnet18.py
```diff
@@ -51,9 +51,10 @@ def forward(self, x):
             s = i % self.frame_per_seq
             x2[i // self.frame_per_seq, s * x.shape[1]:(s + 1) * x.shape[1], :, :] = x[i, :, :, :]
 
-        x = x2
         if x.device.type.startswith('cuda'):
-            x = x.cuda()
+            x = x2.to(x.device)
+        else:
+            x = x2
```
frgfm: You can shorten that to 
`x = x2.to(x.device)` since the device can be a CPU as well ;)

MateoLostanlen: indeed :)

pyronear/datasets/wildfire/split_strategy.py
MateoLostanlen: Done !

MateoLostanlen:
> Thanks again! Would you mind opening a different PR for the model?

Ok, I have created a new PR #69 for that

x0s (changes requested):
Thanks for the PR!

There is quite some duplicate code with already existing features in the repository, so that's is not easy to elicit your contribution and understand what you are achieving here. Usually, when I see myself duplicating lots of code, either I try to refactor as much as possible and/or I look for advice from other contributors in acquaintance with this part of the code. I would be pleased to help, and I am, let's go ;)

As far as I understand what you are achieving here, it seems you are trying to sub-sample a dataset generated with a FrameExtractor which extracted **frames per state** in order to get a dataset based on **frames per sequence**.

Ideas to improve the PRs
- The type of the dataset (`WildFireDataset`) does not seem to change, so it may be unneeded to create a new one(`wildfireSubSampler`) but instead create an **instance of WildFireDataset representing a subsample of wildfire**. In this regards, maybe we should introduce a selector on a 'raw dataset', basically a `metadata.csv` that would be able to _subsample_ on sequences.
- On the other hand, why not enabling the FrameExtractor to extract frames on sequences instead of states ?

Besides, it could be great to comply with the [PEP8 naming convention](https://www.python.org/dev/peps/pep-0008/#naming-conventions). Using [this package](https://github.com/PyCQA/pep8-naming) may help.

Looking forward to improving this contribution along with you,

Cheers,

MateoLostanlen:
@x0s thank you for your comment. I didn't want to pollute WildFireDataSet with things that we may not keep because it's a simple research track. Moreover, I think it's better to keep the WildFireDataSet simple as it is today to be able to build on it. 
Concerning the duplication of code, I totally agree I made myself the remark especially for the WildFireSplitter. I wanted to propose this PR to introduce my idea first and then find a way to generalize WildFireSplitter in a second time. If you think it's better to do it directly in this PR we can do it. I suggest that we keep two different dataset, take the WildFireSplitter out of the wildfire.py file and create an independent class that can be called with a WildFireDataSet or a WildFireSubSamplerDataSet. What do you think?

MateoLostanlen:
PS : Ok for the PEP8, I will change that

x0s:
> @x0s thank you for your comment. I didn't want to pollute WildFireDataSet with things that we may not keep because it's a simple research track. Moreover, I think it's better to keep the WildFireDataSet simple as it is today to be able to build on it.
> Concerning the duplication of code, I totally agree I made myself the remark especially for the WildFireSplitter. I wanted to propose this PR to introduce my idea first and then find a way to generalize WildFireSplitter in a second time. If you think it's better to do it directly in this PR we can do it. I suggest that we keep two different dataset, take the WildFireSplitter out of the wildfire.py file and create an independent class that can be called with a WildFireDataSet or a WildFireSubSamplerDataSet. What do you think?

I agree on keeping things simple. Here it seems possible.
A `class` defines a type, therefore the `WildFireDataset` class defines a type of Dataset. By subsampling the dataset you don't change the type(referring to your code).
Here, you only need to instantiate another `object` with the `WildFireDataset` fed with a `metadata.csv` reflecting the subsampling you need in your research track.

Code may look like this:
```python

metadata_subsampled = compute_subset('metadata.csv')
wildfire_subsampled = WildFireDataset(metadata_subsampled, path_to_frames, target_names=['fire'])

# afterprocessing doesn't change
ratios = {'train': 0.7, 'val': 0.15, 'test':0.15}
splitter =  WildFireSplitter(ratios)
splitter.fit(wildfire_subsampled)
```
This way, this PR would boil down to introducing a function to subsample a metadata file.
What do you think ?

MateoLostanlen:
Hi @x0s , sorry for the very long delay in getting that answer. In order to add a bit of data augmentation I change my sequences at each epoch. If I reduce my PR to a single function then it must be part of the WildFireDataset class. If you think this is still the best solution I can do that :)

x0s:
> If I reduce my PR to a single function then it must be part of the WildFireDataset class.

Actually, you can define such function in the module, there's no need to attach it to `WildFireDataset` class. This way, it will allow subsampling on other `Dataset`.

MateoLostanlen:
Ok I will update the PR, thanks :)

MateoLostanlen:
@x0s I have made the changes, tell me it's ok for you ?

pyronear/datasets/wildfire/wildfire.py
```diff
@@ -5,6 +5,8 @@
 import numpy as np
 import pandas as pd
 import torch
+from random import SystemRandom
```
x0s, line 8: Why are you using `SystemRandom` ? it generates [unreproducible sequences that also change depending on the system](https://docs.python.org/3/library/random.html#random.SystemRandom)

pyronear/datasets/wildfire/wildfire.py
```diff
@@ -191,3 +193,97 @@ def set_splits(self, dataframes):
...
+
+    Parameters
+    ---------
+    metadata: Pandas.DataFrame
+        metadata of the WilFireDataset
+
+    frame_per_seq: int
+        frame per sequence to take
+
+    frame_per_seq: float
```
x0s: typo

pyronear/datasets/wildfire/wildfire.py
```diff
@@ -191,3 +193,97 @@ def set_splits(self, dataframes):
...
+        between 0 and 1, percentage of non-fire data to take twice in order to equalize the dataset
+
+    Example
+    -------
+    subsampler = WildFireSubSampler(dataset.metadata, 2)
+    dataset.metadata = subsampler.computeSubSet()
```
x0s: The `metadata` attribute is not expected to be modified this way.
It is way better to initialize a `Dataset` with the expected metadata instead of modifying the metadata attribute.
Remember that at init, a Dataset is bound to a `metadata` file and a `path leading to the frames`  described in the metadata file.
Dodging init may create divergence and unexpected behaviour

pyronear/datasets/wildfire/wildfire.py
```diff
@@ -191,3 +193,97 @@ def set_splits(self, dataframes):
...
+
+    """
+    def __init__(self, metadata, frame_per_seq, probTh=None):
+
+        self.metadata = metadata
+        self.metadata.index = np.arange(len(self.metadata))
+        self.imgs = self.metadata['imgFile']
+        self.SubSetImgs = []
```
x0s: Please follow PEP8

test/test_datasets_wildfire_subsampler.py
```diff
@@ -0,0 +1,55 @@
...
+        self.wildfire_path = Path(__file__).parent / 'fixtures/subsampler.csv'
+        self.wildfire_df = pd.read_csv(self.wildfire_path)
+
+    def test_good_size_after_subsamping(self):
+        wildfire = WildFireDataset(metadata=self.wildfire_path,
+                                   path_to_frames=self.path_to_frames)
+
+        self.assertEqual(len(wildfire), 1999)
+
+        subsampler = WildFireSubSampler(wildfire.metadata, 2)
```
x0s: The Subsampler input is a metadata file and outputs another metadata file.

Why not simply init a WildFireDataset with the new subsampled metadata file?

test/fixtures/subsampler.csv
```diff
@@ -0,0 +1,2000 @@
+,exploitable,fire,sequence,clf_confidence,loc_confidence,x,y,t,stateStart,stateEnd,imgFile,fire_id
```
x0s: Why are you completing already existing fixtures ? (https://github.com/pyronear/PyroNear/blob/master/test/fixtures/wildfire_states.csv)

pyronear/datasets/wildfire/wildfire.py
```diff
@@ -191,3 +193,97 @@ def set_splits(self, dataframes):
         # Determine estimated(posterior) parameters
         self.n_samples_ = {set_: len(self.splits[set_]) for set_ in ['train', 'val', 'test']}
         self.ratios_ = {set_: (self.n_samples_[set_] / len(self.wildfire)) for set_ in ['train', 'val', 'test']}
+
+
+class WildFireSubSampler:
```
x0s: Actually this subsampler is not bound to WildFire Dataset and can be used with any wildfire video dataset annotated the same way. To help user understand what's made here, The class name can reflect better this possibility

pyronear/datasets/wildfire/wildfire.py
```diff
@@ -191,3 +193,97 @@ def set_splits(self, dataframes):
...
+    """Sub-Sample a WildFireDataset
+
+    The Wildfire DataSet is annotated with video sequences. To extract this dataset as frames,
+    you have to choose the number N of frames per sequence. The idea of this SubSampler is to
+    combine K frames among these N to create a derivative of the resnet the ssresnet.
+
+    Parameters
+    ---------
+    metadata: Pandas.DataFrame
```
x0s, line 203: It could be great to share [same interface](https://github.com/pyronear/PyroNear/blob/master/pyronear/datasets/wildfire/wildfire.py#L24) with `WildFireDataset`

test/test_datasets_wildfire_subsampler.py
```diff
@@ -0,0 +1,55 @@
...
+        wildfire = WildFireDataset(metadata=self.wildfire_path,
+                                   path_to_frames=self.path_to_frames)
+
+        subsampler = WildFireSubSampler(wildfire.metadata, 2)
+        wildfire.metadata = subsampler.computeSubSet()
+
+        meta2 = subsampler.computeSubSet()
+
+        self.assertEqual(len(meta2), 400)
+        self.assertFalse(wildfire.metadata['imgFile'].values.tolist() == meta2['imgFile'].values.tolist())
```
x0s: We definitely need reproducibility when dealing with random numbers. If not ensured, we cannot guarantee any result nor stability.
If you don't want to be bound to a random generator, seeds can be hyperoptimized, while maintaining the predictor deterministic.

test/test_datasets_wildfire_subsampler.py
```diff
@@ -0,0 +1,55 @@
...
+# See the LICENSE file in the root of this repository for complete details.
+
+import unittest
+from pathlib import Path
+import pandas as pd
+
+from pyronear.datasets.wildfire import WildFireDataset, WildFireSubSampler
+
+
+class WildFireDatasetSubSampler(unittest.TestCase):
```
x0s: Please append with `Tester` so this class name doesn't collide with others

pyronear/datasets/wildfire/wildfire.py
```diff
@@ -191,3 +193,97 @@ def set_splits(self, dataframes):
...
+    """
+    def __init__(self, metadata, frame_per_seq, probTh=None):
+
+        self.metadata = metadata
+        self.metadata.index = np.arange(len(self.metadata))
+        self.imgs = self.metadata['imgFile']
+        self.SubSetImgs = []
+        self.probTh = probTh
+        self.frame_per_seq = frame_per_seq
+        # Define sequences numbers
```
x0s: Why are you re-extracting the fBase from the imgFile ? Shouldn't it be already present since previous step was Frame extraction  ?

x0s (dismissed):
Hi Thanks for adapting your code,

You seem to use a class where only a function is needed. There're actually not benefit doing so, because it encourage implicit states. As you know from the zen of python:
```
Explicit is better than implicit.
Simple is better than complex.
```

I don't have time to review the algorithm(subsampling part), but it could be great to have more comments to understand how you make it.

Also, we may use a lighter fixture for the `checkedlabels` or `metadata`. It could be good to make it consistent with already existing fixtures to avoid confusion.

Deterministic behaviour is really expected, we need to be able to reproduce our result so they're stable across time

Also please make the whole PR compliant to [PEP8 naming convention](https://www.python.org/dev/peps/pep-0008/#naming-conventions) as you said it to do so:
> PS : Ok for the PEP8, I will change that

Using [this package](https://github.com/PyCQA/pep8-naming) may help.

pyronear/datasets/wildfire/wildfire.py
MateoLostanlen: You are saying that because the s of SubSetImgs should be lower case right ?

pyronear/datasets/wildfire/wildfire.py
MateoLostanlen: Yes, you are right sorry for that

test/test_datasets_wildfire_subsampler.py
MateoLostanlen: I'm sorry but I don't understand your comment :/

test/test_datasets_wildfire_subsampler.py
MateoLostanlen: You are right, I have change that !

pyronear/datasets/wildfire/wildfire.py
MateoLostanlen: When I use random I got this error
 `Standard pseudo-random generators are not suitable for security/cryptographic purposes.` 
in Codacy/PR Quality Review so I switch to SystemRandom

pyronear/datasets/wildfire/wildfire.py
MateoLostanlen: What is the issue here ?

test/test_datasets_wildfire_subsampler.py
MateoLostanlen: Done !

pyronear/datasets/wildfire/wildfire.py
frgfm: I think he was referring to the double parameter's description of `frame_per_seq`!

test/test_datasets_wildfire_subsampler.py
frgfm: I think he was suggesting to append `Tester` at the end of this class's name :) (so `WildFireDatasetSubSamplerTester` or something like `WildFireSubSamplerTester` is you want a shorter name)

pyronear/datasets/wildfire/wildfire.py
MateoLostanlen: Not relevent anymore with last update

pyronear/datasets/wildfire/wildfire.py
MateoLostanlen: Done

pyronear/datasets/wildfire/wildfire.py
MateoLostanlen: Ok thanks, not here anymore in last update

pyronear/datasets/wildfire/wildfire.py
MateoLostanlen: Thanks, it's fixed

test/fixtures/subsampler.csv
MateoLostanlen: It's an error on my side, I don't not why it wasn't in my metadata anymore

pyronear/datasets/wildfire/wildfire.py
frgfm: I'm not familiar with `SystemRandom` but if we only need a scalar random value, you can easily discard Codacy warning. If there is indeed a lack of availability on some systems as mentioned in the earlier reference, it might be safer to stick with a bare `random.random()`.

Alternatively, this file has a numpy dependency which has a large support of `numpy.random` !

pyronear/datasets/wildfire/wildfire.py
frgfm: Since the files changed, is this still an issue?

pyronear/datasets/wildfire/wildfire.py
frgfm: Not sure if we were talking about the same thing but there are two mentions of the `frame_per_seq`, and there is no mention of the `seed`argument in the constructor docstring. Mind fixing that up? :)

test/test_datasets_wildfire_subsampler.py
frgfm: Would you mind changing the class name so that we can close this thread? :pray:

frgfm (changes requested):
Apart from the docstring to fix, the PR looks alright, almost there! I'm still having concerns about adding new `.csv` content files each time we want to add a new feature and its unittests though.

MateoLostanlen:
> 
> 
> Apart from the docstring to fix, the PR looks alright, almost there! I'm still having concerns about adding new `.csv` content files each time we want to add a new feature and its unittests though.

We do not need a new csv for each feture only a  csv complete enough to test all cases. I suggest to open an issue on this subject and to create a PR soon to solve it.

I just made a commit to change the name of the test class and to remove the error in the doc. Can we validate the PR?

frgfm:
> > Apart from the docstring to fix, the PR looks alright, almost there! I'm still having concerns about adding new `.csv` content files each time we want to add a new feature and its unittests though.
> 
> We do not need a new csv for each feture only a csv complete enough to test all cases. I suggest to open an issue on this subject and to create a PR soon to solve it.
> 
> I just made a commit to change the name of the test class and to remove the error in the doc. Can we validate the PR?

Just missing the `seed` argument in the docstring :pray:
Well, it's far from being one .csv for all features. In this PR for instance, we introduce a 2000 lines content file for testing. Let's discuss this in an issue as you mentioned, it will be easier then!

frgfm (dismissed):
Looks good to me! I'll open an issue later for fixture & content files addition in PR, thanks!

frgfm (approved):
Looks good to me!

PR #67, opened 2020-05-18
chore: Cleaned setup.py and updates gitignore

frgfm:
This PR cleans setup.py of an unused import and adds `/data` in gitignore. 

EDIT: added a setup.cfg for future conda release.


x0s:
Hi,
You didn't mention what are you dealing with conda.
Also, you seem to link a LICENSE that no longer applies since we are now contributing as a team.

frgfm:
> Hi,
> You didn't mention what are you dealing with conda.
> Also, you seem to link a LICENSE that no longer applies since we are now contributing as a team.

Correct, I pushed the last commit later, sorry!
We have to update the License as well but the modification in this PR is still valid, isn't it?
(it only points to the License)

x0s:
> > Hi,
> > You didn't mention what are you dealing with conda.
> > Also, you seem to link a LICENSE that no longer applies since we are now contributing as a team.
> 
> Correct, I pushed the last commit later, sorry!
> We have to update the License as well but the modification in this PR is still valid, isn't it?
> (it only points to the License)

Why would we want to release on Conda an invalid License ?

frgfm:
> Why would we want to release on Conda an invalid License ?

Hey there,
I can change my default name to Pynear Contributors if you like, but this seems like a topic for a different PR. Or I can remove the setup.cfg.

Again, here I just wanted to configure everything to point to the updated version of files required to publish. I'm just suggesting to set up a "pointer" here, changing the content of the License is fine but feels like something for a different PR. Besides, I think we still have an ongoing discussion about which license to use.

What do you think is best? Should I remove the `setup.cfg`? or edit my name in the license into Pyronear contributors?

x0s:
I think things should get done in reversed order:
We choose a valid license, then we release the lib **pointing to it** not the other way around.

frgfm:
> I think things should get done in reversed order:
> We choose a valid license, then we release the lib **pointing to it** not the other way around.

No worries, just removed `setup.cfg`!
Let's open up a PR later specifically for the license

x0s (approved):

PR #69, opened 2020-05-22
Add sub sampler resnet18

MateoLostanlen:
The goal of this PR is to add the SubSamplerResnet18 model which is a derivative of the resnet18 that we want to train on the SubSamplerDataSet of the PR #66


pyronear/models/ssresnet18.py
```diff
@@ -0,0 +1,78 @@
...
+        self.load_state_dict(dictres18)
+
+        self.frame_per_seq = frame_per_seq
+        self.fc = nn.Linear(256, 1)
+        self.conv1_1 = nn.Conv2d(512 * self.frame_per_seq, 512, kernel_size=1)
+        self.conv3_1 = nn.Conv2d(512, 512, kernel_size=3)
+        self.conv1_2 = nn.Conv2d(512, 256, kernel_size=1)
+        self.conv2_bn = nn.BatchNorm2d(512)
+        self.conv2_bn1 = nn.BatchNorm2d(512)
+        self.conv2_bn2 = nn.BatchNorm2d(256)
```
frgfm: it would be cleaner to make a `self.layer5` here ;)

pyronear/models/ssresnet18.py
```diff
@@ -0,0 +1,78 @@
+from torchvision.models.resnet import ResNet, BasicBlock
+import torchvision
+from torch import nn
+import torch
+import torch.nn.functional as F
+
+
+class SSResNet18(ResNet):
+    def __init__(self, frame_per_seq):
+        super(SSResNet18, self).__init__(BasicBlock, [2, 2, 2, 2])
```
frgfm: For your model to be easily reusable, I would suggest rather creating a `SSResNet` class, and have a `ssresnet18` function that will return the correct layer layout (similarly to what is done in [torchvision](https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L232)

pyronear/models/ssresnet18.py
```diff
@@ -0,0 +1,78 @@
...
+        nn.init.kaiming_normal_(self.conv1_2.weight, mode='fan_out', nonlinearity='relu')
+
+        nn.init.constant_(self.conv2_bn.weight, 1)
+        nn.init.constant_(self.conv2_bn.bias, 0)
+
+        nn.init.constant_(self.conv2_bn1.weight, 1)
+        nn.init.constant_(self.conv2_bn1.bias, 0)
+
+        nn.init.constant_(self.conv2_bn2.weight, 1)
+        nn.init.constant_(self.conv2_bn2.bias, 0)
```
frgfm: Each layer type shares the same initialization, try to consider doing something like this instead: https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L160

Since you would be moving the pretrained option out of the class to the function, this would be a better option ;)

frgfm (commented):
Great PR! I made a few suggestions to get this to be easily reusable on other resnet-based achitecture, let me know what you think!

MateoLostanlen:
Hi @frgfm , Thanks for your remarks. It is indeed a good idea to generalize to any resnet. Let me know what you think of this new version

pyronear/models/ssresnet.py
```diff
@@ -0,0 +1,122 @@
...
+                Output shape of the first conv1x1
+
+            output2L5: int
+                Output shape of the second conv1x1
+
+            """
+        if block == 'BB':
+            super(SSResNet, self).__init__(BasicBlock, layers)
+        else:
+            super(SSResNet, self).__init__(Bottleneck, layers)
```
frgfm: You're making it harder for you :) 
just do `super().__init__(block, layers)` without conditions

and add function for each resnet size just like [here](https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L240)

frgfm: And pretrained weight loading would be better in the function that in the model definition :+1:

frgfm (commented):
Almost there! you can define each ssresnet size like in torchvision and move weight loading outside of the class definition, check my comments, and we'll be good to go!

pyronear/models/ssresnet.py
MateoLostanlen: I just committed the changes, based on torchvision. I'll open a new PR soon to generalize to all resnets and handle the import of a pre-trained model.

pyronear/models/ssresnet.py
```diff
@@ -1,64 +1,49 @@
-from torchvision.models.resnet import ResNet, BasicBlock, Bottleneck
+from torchvision.models.resnet import ResNet, resnet18, BasicBlock
 import torchvision
 from torch import nn
 import torch
 import torch.nn.functional as F
 
 
+__all__ = ["SSResNet"]
```
frgfm, line 6: ssresnet18 missing here, isn't it? :)

pyronear/models/ssresnet.py
```diff
@@ -0,0 +1,115 @@
...
+    Then we need shapes of the layer5
+    intput1L5: int
+        Input of the layer5, must be equal to the output of layer4
+    output1L5: int
+        Output shape of the first conv1x1
+    output2L5: int
+        Output shape of the second conv1x1
+    """
+    def __init__(self, block, layers, frame_per_seq=2, intput1L5=512,
+                 output1L5=512, output2L5=256):
```
frgfm: Not exactly sure, but the `512` is actually here because the last conv layer of resnet18 has 512 channels. If so, I'd suggest to retrieve dynamically that value using this https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L158

It's actually `512 * block.expansion`, so you don't need `input1L5` or `output1L5`. Regarding `output2L5`, perhaps renaming it into something more user-friendly would be beneficial ;)

frgfm (commented):
Thanks for the PR! This already looks good, I made a few suggestions to improve flexibility, let me know your thoughts on it!

pyronear/models/ssresnet.py
MateoLostanlen: No, I removed it on purpose. Otherwise this test fail : https://github.com/pyronear/PyroNear/blob/ec3ba3116135d853e5eb10836ba509c9e04462bc/test/test_models.py#L62

I will put it back when a pretrained model will be available

pyronear/models/ssresnet.py
MateoLostanlen: Ok tanhks for that, I have fixed it :)

pyronear/models/ssresnet.py
frgfm: Good catch! But then, you might have to remove it from the documentation?

test/test_models.py
```diff
@@ -3,6 +3,8 @@
 import numpy as np
 import random
 from pyronear import models
+import torchvision
```
frgfm: Codacy seems to say that this is an unused import, you can remove it ;)

pyronear/models/ssresnet.py
```diff
@@ -0,0 +1,106 @@
+from torchvision.models.resnet import ResNet, resnet18, BasicBlock
+import torchvision
+from torch import nn
+import torch
+import torch.nn.functional as F
```
frgfm: Codacy seems to say that `resnet18`, `torchvision`, `import torch.nn.functional as F` an unused imports, you can remove it ;) (cf. the codacy check)

frgfm (commented):
A few minor adjustments for the final touch, and we're good to go!

pyronear/models/ssresnet.py
MateoLostanlen: Yes maybe it's better for know

test/test_models.py
MateoLostanlen: ok thanks

frgfm (approved):
Thanks a lot, looking forward to try it out!

PR #71, opened 2020-05-26
Update project setup documentation to clarify virtualenvwrapper init (fix #70)

x0s:
This PR aims at updating the documentation to clarify environment setup and avoid python 2/3 conflict. Therefore it resolves #70

PR #68, opened 2020-05-18
chore: Switched from CircleCI to GitHub workflow

frgfm:
This PR aims at changing CI from third-party CircleCI to Github workflow by:
- moving previous CI jobs to Github workflow
- adding a workflow to check documentation build status
- adding a workflow for documentation deployment

_Please note that CircleCI legacy checks are expected to fail since they are not run anymore_

PR #73, opened 2020-05-26
chore: Fixes follow-ups issues from #68

frgfm:
This PR fixes some issues raised by @x0s in #68 by:
- setting back the Python version support from 3.7 to 3.6
- adding back the flake8 requirement for local testing
- triggering workflow on pull request and not commit to fix non-running workflows on #71 

Any feedback is welcome!

### DIFF
--- a/.github/workflows/doc-deploy.yml
+++ b/.github/workflows/doc-deploy.yml
@@ -10,10 +10,10 @@ jobs:
       - uses: actions/checkout@v2
         with:
           persist-credentials: false
-      - name: Set up Python 3.7
+      - name: Set up Python 3.6
         uses: actions/setup-python@v1
         with:
-          python-version: 3.7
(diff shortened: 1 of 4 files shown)


x0s (approved):
Thanks for taking my comments into consideration. 
I am not against moving up the python version and benefit from the great new strenghts, but first we may assess consequences and communicate clearly with contributors/users.

Do you have any clues why building is still failing ?

frgfm:
> Do you have any clues why building is still failing ?

I'm not sure what you are asking exactly?
As mentioned previously, CircleCI builds are expected to fail. Regarding github workflows, I suppose it was because the trigger event had to be forced to pull_request, hence the suggestion in this PR.

To know for sure, I'd suggest you make the change on one of your PRs, and if it works, merge master into your branch I guess :)

PR #74, opened 2020-05-28
chore: Ensure squashed merged PR triggers workflow

frgfm:
Currently, PR trigger the main workflow, but once the commits are squash-merged into master, the workflow do not run again.

This PR enforces workflow trigger on squash merged commits into master.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,9 @@
 name: python-package
 
 on:
-  pull_request
+  pull_request:
+    branches:
+      - master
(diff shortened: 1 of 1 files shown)


frgfm:
@blenzi you might want to do the same thing for the repos you recently set up CI on ;)

blenzi (approved):
> @blenzi you might want to do the same thing for the repos you recently set up CI on ;)

Thanks @frgfm! This is already the case for VideoSplitter and VideoAnnotator

PR #76, opened 2020-06-01
refactor: Cleaned code base

frgfm:
This PR cleans code base by:
- removing unused imports
- removing unused variables
- switching from `assert` to `AssertionError` syntax

_Note: these modifications were spotted by codacy_

### DIFF
--- a/pyronear/datasets/wildfire/fire_labeler.py
+++ b/pyronear/datasets/wildfire/fire_labeler.py
@@ -109,7 +109,8 @@ class FireLabeler:
 
         length = len(self.fire_ids[self.fire_ids == -1])
         self.fire_ids[self.fire_ids == -1] = range(current_fire_id, current_fire_id + length)
-        assert self.fire_ids[self.fire_ids == -1].sum() == 0, "Singletons escaped indexation!!"
+        if self.fire_ids[self.fire_ids == -1].sum() != 0:
+            raise AssertionError("Singletons escaped indexation!")
         return self
 
     @staticmethod
--- a/pyronear/datasets/wildfire/fire_labeler.py
+++ b/pyronear/datasets/wildfire/fire_labeler.py
@@ -109,8 +109,7 @@ class FireLabeler:
 
         length = len(self.fire_ids[self.fire_ids == -1])
         self.fire_ids[self.fire_ids == -1] = range(current_fire_id, current_fire_id + length)
-        if self.fire_ids[self.fire_ids == -1].sum() != 0:
-            raise AssertionError("Singletons escaped indexation!")
+        assert self.fire_ids[self.fire_ids == -1].sum() == 0, "Singletons escaped indexation!"
         return self
 
     @staticmethod
--- a/pyronear/datasets/wildfire/frame_extractor.py
+++ b/pyronear/datasets/wildfire/frame_extractor.py
@@ -207,7 +207,7 @@ class FrameExtractor:
             # Get the video
             movie = cv2.VideoCapture((self.path_to_videos / name).as_posix())
             # For each state
-            for index, row in group.iterrows():
+            for _, row in group.iterrows():
                 # Position the video at the current frame
                 movie.set(cv2.CAP_PROP_POS_FRAMES, row.frame)
                 success, frame = movie.read()
--- a/pyronear/datasets/wildfire/frame_extractor.py
+++ b/pyronear/datasets/wildfire/frame_extractor.py
@@ -207,7 +207,7 @@ class FrameExtractor:
             # Get the video
             movie = cv2.VideoCapture((self.path_to_videos / name).as_posix())
             # For each state
-            for _, row in group.iterrows():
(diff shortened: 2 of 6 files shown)


pyronear/datasets/wildfire/fire_labeler.py
```diff
@@ -109,7 +109,8 @@ def run(self):
 
         length = len(self.fire_ids[self.fire_ids == -1])
         self.fire_ids[self.fire_ids == -1] = range(current_fire_id, current_fire_id + length)
-        assert self.fire_ids[self.fire_ids == -1].sum() == 0, "Singletons escaped indexation!!"
+        if self.fire_ids[self.fire_ids == -1].sum() != 0:
+            raise AssertionError("Singletons escaped indexation!")
```
x0s: Assertions are here to help developers to debug errors that should never happen, not to halt the program at runtime. Thence, that's misleading to rewrite it the way Exceptions are raised in the program. Indeed, AssertionErrors are **never expected** to be caught programmatically.

Here, the code suggested in this PR seems not to change how the Assertion is raised, but:
- it does lessen the clarity how to address it (treat like an Exception where it shouldn't)
- it prevents code optimization at runtime by ignoring ` __debug__` (see [doc](https://docs.python.org/3.6/reference/simple_stmts.html#the-assert-statement))

pyronear/datasets/wildfire/frame_extractor.py
```diff
@@ -207,7 +207,7 @@ def _write_frames(self, labels: pd.DataFrame, path_to_frames: Union[str, Path])
             # Get the video
             movie = cv2.VideoCapture((self.path_to_videos / name).as_posix())
             # For each state
-            for index, row in group.iterrows():
+            for _, row in group.iterrows():
```
x0s: Here, It depends what we want:
- Using conventional unindentified `_` to tell the developer **that this variable won't be used**
- Using explicit name `index` to tell the developer **how this variable could be used**

Look good to me either way.

x0s (changes requested):
Thanks for the PR,
- Good to remove unused imported libraries.
- `Assertion` behavior change may be not advisable.

pyronear/datasets/wildfire/frame_extractor.py
frgfm: Agreed, I dug into the different options here and switched to `.itertuples`, which has the same behavior (without yielding indices) and expected to have a faster execution.

One could also use  `apply` syntax, I tried and it works but I feel like we would lose some code clarity in doing this.

pyronear/datasets/wildfire/fire_labeler.py
frgfm: Oh I see! But then I don't see whether this is needed in this case? If we decide to process this case in the future, we can simply remove the `AssertionError` (or put it in an `else` case), just like we would have to with `assert`? _(I've already seen several references to [this](https://docs.openstack.org/bandit/latest/plugins/b101_assert_used.html) indicating that developers should switch to `AssertionError` hence my suggestion)_

I don't know how this indexation case serious is actually. If we only need to tell the user without ever halting runtime, perhaps we could switch to warnings instead?

Also, I'm curious but what do you mean about code optimization? If I read the doc correctly, the code optimization will be different only if `__debug__` is set to `False`, and in this case, the `assert` will be removed (meaning that this case will not be processed). Perhaps it's expected, as I honestly don't know the importance of this indexation case :thinking: 

Let me know what you think, you know the use case much better than I do :)

pyronear/datasets/wildfire/frame_extractor.py
x0s: Indeed, `itertuples` is a better option!

pyronear/datasets/wildfire/fire_labeler.py
x0s: Using `assert` is a bad practice if one uses it in case of protecting code that can fail at runtime (in this case, we raise Exceptions)
OpenStack decided to prevent every use of assert because it seems that their contributors were using it too much _enforce interface constraints_ (wrong way, tests are here for this)

Here that's not the case, the `assert` is here to say: _at this point of the algorithm, the state of this Series must be **totally** deprived of unassigned items_ 




It warns future contributors that this is important to maintain the state of this Series as such. Because it can generate hard-to-debug errors. That's why we want the program to halt in dev mode but not in production.
`warnings` are not suitable because they can be ignored, here we want the program to halt quickly, so the contributor can directly dive into the code and fix the issue. It is particularly handy when refactoring algorithms.

Concerning optimization, we want to get rid of every stuff solely dedicated to dev mode when in production. This assertion is absolutely unneeded in production, so it can be deleted.

frgfm: Fair enough, thanks for the explanation about the expected behaviour here!
I just reverted back to `assert`

x0s (approved):

PR #77, opened 2020-06-03
fix: making sure that path_to_frames are path type and not str

### LINKED ISSUE #59: [datasets] Documentation error within WildfireDataset class where path_to_frame should be a path type
## 🐛 Bug

`path_to_frame` should be a path type and not an str as mentioned below -->

https://github.com/frgfm/PyroNear/blob/61d5d527187a6f6b0696da0bc199c1a80a2ee2c0/pyronear/datasets/wildfire/wildfire.py#L31-L32

str is not supported by PyTorch when iterating trough DataLoader

## To Reproduce

1. str path 

`path_to_frames = "/Users/sebastienelmaleh/Google Drive/Deep Learning/Pyronear/frames_raw_wildfire/"`

2. Iterating through DataLoader 

```
for epoch in range(num_epochs):
    for i, (images, labels) in enumerate(wildfire_loader_train):
```
which produces : 

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-18-4d31344fa390> in <module>
      4 iter = 0
      5 for epoch in range(num_epochs):
----> 6     for i, (images, labels) in enumerate(wildfire_loader_train):
      7 
      8         labels = labels.squeeze()

/usr/local/lib/python3.7/site-packages/torch/utils/data/dataloader.py in __next__(self)
    344     def __next__(self):
    345         index = self._next_index()  # may raise StopIteration
--> 346         data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
    347         if self._pin_memory:
    348             data = _utils.pin_memory.pin_memory(data)

/usr/local/lib/python3.7/site-packages/torch/utils/data/_utils/fetch.py in fetch(self, possibly_batched_index)
     42     def fetch(self, possibly_batched_index):
     43         if self.auto_collation:
---> 44             data = [self.dataset[idx] for idx in possibly_batched_index]
     45         else:
     46             data = self.dataset[possibly_batched_index]

/usr/local/lib/python3.7/site-packages/torch/utils/data/_utils/fetch.py in <listcomp>(.0)
     42     def fetch(self, possibly_batched_index):
     43         if self.auto_collation:
---> 44             data = [self.dataset[idx] for idx in possibly_batched_index]
     45         else:
     46             data = self.dataset[possibly_batched_index]

~/seb_PyroNear/PyroNear/pyronear/datasets/wildfire/wildfire.py in __getitem__(self, index)
     59             - x,y (float, float) and
     60             - Exploitable(True/False)"""
---> 61         path_to_frame = self.path_to_frames / self.metadata['imgFile'].iloc[index]
     62         observation = self.load_image(path_to_frame)
     63 

TypeError: unsupported operand type(s) for /: 'str' and 'str'

## Expected behavior

Importing pathlib and converting str to path type corrected the error

Akilditu:
This PR aims at fixing the path_to_frames type that used to be mentioned as str but needs to be a path. 

Fixes #59

### DIFF
--- a/pyronear/datasets/wildfire/wildfire.py
+++ b/pyronear/datasets/wildfire/wildfire.py
@@ -29,7 +29,7 @@ class WildFireDataset(Dataset, VisionMixin):
         If left to None, will be set to ['fire']
         Example: ['fire', 'clf_confidence', 'loc_confidence', 'x', 'y']
 
-    path_to_frames: str or path 
+    path_to_frames: str or path
         Path leading to the directory containing the frames referenced in metadata 'imgFile':
 
     transform: object, optional
--- a/pyronear/datasets/wildfire/wildfire.py
+++ b/pyronear/datasets/wildfire/wildfire.py
@@ -6,7 +6,7 @@ import numpy as np
 import pandas as pd
 import torch
 from torch.utils.data import Dataset
-from pathlib import Path, PurePath
+from pathlib import Path
 
 from ..utils import VisionMixin
 from .split_strategy import ExhaustSplitStrategy
@@ -29,7 +29,7 @@ class WildFireDataset(Dataset, VisionMixin):
         If left to None, will be set to ['fire']
         Example: ['fire', 'clf_confidence', 'loc_confidence', 'x', 'y']
 
-    path_to_frames: path type
+    path_to_frames: str or path 
         Path leading to the directory containing the frames referenced in metadata 'imgFile':
 
     transform: object, optional
@@ -49,11 +49,7 @@ class WildFireDataset(Dataset, VisionMixin):
         self.transform = transform
 
         # converting path_to_frames to path type
-
-        if isinstance(path_to_frames, PurePath):
-            self.path_to_frames = path_to_frames 
-        else :
-            self.path_to_frames = Path(path_to_frames)
+        self.path_to_frames = Path(path_to_frames)
 
     def __len__(self):
         return len(self.metadata)
--- a/pyronear/datasets/wildfire/wildfire.py
+++ b/pyronear/datasets/wildfire/wildfire.py
@@ -6,6 +6,7 @@ import numpy as np
 import pandas as pd
 import torch
 from torch.utils.data import Dataset
+from pathlib import Path, PurePath
 
 from ..utils import VisionMixin
 from .split_strategy import ExhaustSplitStrategy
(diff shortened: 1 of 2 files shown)


pyronear/datasets/wildfire/wildfire.py
```diff
@@ -28,7 +29,7 @@ class WildFireDataset(Dataset, VisionMixin):
         If left to None, will be set to ['fire']
         Example: ['fire', 'clf_confidence', 'loc_confidence', 'x', 'y']
 
-    path_to_frames: str
+    path_to_frames: path type
```
x0s: It could be great to ensure retro-compatibility and flexibility: being able to accept `strings` and `Path-like` objects!

pyronear/datasets/wildfire/wildfire.py
```diff
@@ -45,9 +46,15 @@ def __init__(self, metadata, path_to_frames, target_names=None, transform=None):
 
         # default target is fire detection (0/1)
         self.target_names = target_names or ['fire']
-        self.path_to_frames = path_to_frames
         self.transform = transform
 
+        # converting path_to_frames to path type
+
+        if isinstance(path_to_frames, PurePath):
```
x0s: Python `Path` object is idempotent over strings, so you can directly use it either on strings or paths:
```python
path_str = "/path/to/lovely/directory"
path_pure = Path("/path/to/lovely/directory")

assert Path(path_str) == path_pure # True
assert Path(path_pure) == path_pure # True (idempotence)
```

x0s (changes requested):
Thanks for the PR!
Here are some comments how to improve flexibility (being able to use both strings and path-like objects)

pyronear/datasets/wildfire/wildfire.py
Akilditu: Thank you for the comment but I don't see a use case where strings would be useful ?

pyronear/datasets/wildfire/wildfire.py
Akilditu: Ok then the if statement isn't relevant here, will simplify applying Path directly on input path_to_frames.

pyronear/datasets/wildfire/wildfire.py
x0s: For instance, if you are loading project parameters from config file (such as path to frames), you would get a string. In this case, it would be handy to not have to convert it to a Path-like object but directly pass it to the WildFire class initialisation

Akilditu: Understood, adding changes in next commit

x0s (changes requested):
Thanks taking my remarks in to account,

Could you add some tests that cover initializing a `WildFireDataset` (str and path) so future contributions won't raise again the issue you solved by your work ? Meanwhile, it will make codecoverage check pass.
After this, I think we'll be ready to merge ;)

x0s (approved):

PR #79, opened 2020-06-28
First inference

### LINKED ISSUE #65: Add demo
## 🚀 Feature
I think we should add a demo script that loads a pre-training model and applies it to a video sequence.

## Motivation

The purpose of this demo is to show an example of how to use the latest pyronear model, allowing anyone to use it on the video of their choice.

MateoLostanlen:
I think we're at a stage in the project where we should always have a demo model available. The purpose of this PR is to open this door and propose a first model. This model is very simple and not very efficient. It should be used to make first tests on rapsberry pi and also to encourage us to propose better !


frgfm:
Hey there @MateoLostanlen,

Thanks again! Would you mind adding instructions on how to use your feature in the README for instance?
I'm not exactly sure on how you expect it to be used. And same as in the other PR, I would advise not to put large files (40Mb+ here) in the commits but rather in the release attachments.

Cheers!

MateoLostanlen:
The goal of this PR is to have a kind off demo to use pyronear predictor without a rapsberry pi, I will update the readme soon ;)

MateoLostanlen:
This will be remove and replace by a webApp : https://hub.gke.mybinder.org/user/mateolostanlen-testwebapp-birhzdvj/voila/render/webApp.ipynb?token=kQxhymsFRB-XmgEu-TLLag

PR #80, opened 2020-07-18
Pyronear engine

MateoLostanlen:
This class is the Pyronear Engine. This engine manage the whole Fire Detection process on raspberry pi by capturing and saving the image and by predicting if there is a fire or not based on this image

### DIFF
--- a/raspberryPi/pyronearEngine.py
+++ b/raspberryPi/pyronearEngine.py
@@ -6,7 +6,8 @@ import torch
 from torch import nn
 from PIL import Image
 from torchvision import transforms
-import smtplib, ssl
+import smtplib
+import ssl
 
 
 class PyronearEngine:
--- a/raspberryPi/pyronearEngine.py
+++ b/raspberryPi/pyronearEngine.py
@@ -82,7 +82,7 @@ class PyronearEngine:
         smtp_server = "smtp.gmail.com"
         sender_email = "test.pyronear@gmail.com"  # Enter your address
         receiver_email = "mateo.lostanlen@gmail.com"  # Enter receiver address
-        password = ""  # Add your password
+        # password = ""  # uncomment and add your password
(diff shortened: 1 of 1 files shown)


frgfm:
Thanks @MateoLostanlen for the PR!
It's indeed a great feature, I'll look into the class shortly but I have a question:
- what is the .txt file in the PR? Is this a model checkpoint? If so, isn't it the wrong extension?
- assuming the file is actually more than 40Mb, I would suggest that we put it in the next release attachments, rather than in the source code. What do you think?

MateoLostanlen:
Hi @frgfm, yes the txt file are the weights of the model. Ok I will remove it, if you think it's better this way

frgfm:
> Hi @frgfm, yes the txt file are the weights of the model. Ok I will remove it, if you think it's better this way

For future reference, whether we put the checkpoint in the release attachments or somewhere else, it's better to use the `.pth` or `.pt` extension rather than `.txt` ;) (out of curiosity I tried to open the txt file with my IDE and found out it was a binary file)

MateoLostanlen:
> > Hi @frgfm, yes the txt file are the weights of the model. Ok I will remove it, if you think it's better this way
> 
> For future reference, whether we put the checkpoint in the release attachments or somewhere else, it's better to use the `.pth` or `.pt` extension rather than `.txt` ;) (out of curiosity I tried to open the txt file with my IDE and found out it was a binary file)

ok thanks :)

frgfm:
Just like in #81, I think this PR would be better on the embedded dedicated repo https://github.com/pyronear/PyronearEngine
If so, maybe we can close this PR and follow-up on the later?

MateoLostanlen:
New repo for this topic : https://github.com/pyronear/PyronearEngine

I close the PR

PR #86, opened 2020-12-11
Create pyronear predictor

MateoLostanlen:
The purpose of this PR is to propose an create a Pyronear Predictor. This class will load the last Pyronear model and allow to perform a prediction


pyrovision/inference/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) Pyronear contributors.
+# This file is dual licensed under the terms of the CeCILL-2.1 and AGPLv3 licenses.
+# See the LICENSE file in the root of this repository for complete details.
+
+from .pyronear_predictor import PyronearPredictor
```
frgfm: I think we should only be adding a model in `pyrovision.models`
Just like in torchvision (https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py) we can add the checkpoint URL in a dict at the top of the file, and write a function to instantiate the model (with pretrained as argument)

pyrovision/inference/inference.cfg
```diff
@@ -0,0 +1,13 @@
...
+use_adaptive_concat_pool2d_head = True
+nb_features = 1280
+cut = -2
+url = https://github.com/pyronear/pyro-vision/releases/download/v0.1.0/rexnet1_0x_acp_2e017f83.pth
+device = cuda
+
+[TRANSFORMS]
+image_size = 448
+use_center_crop = True
+
```
frgfm: We cannot afford to add a config here for each model that we will train. But we could pass these params to the function instantitating the model, and if `pretrained=True`, download the checkpoint from the URL & load it

pyrovision/inference/pyronear_predictor.py
```diff
@@ -0,0 +1,106 @@
...
+            tf = transforms.Compose([transforms.Resize(size=(size)),
+                                     transforms.CenterCrop(size=size),
+                                     transforms.ToTensor(),
+                                     normalize
+                                     ])
+        else:
+            tf = transforms.Compose([transforms.Resize(size=(size)),
+                                     transforms.ToTensor(),
+                                     normalize
+                                     ])
```
frgfm: Each user can have his/user own transforms, so perhaps we can let them choose. Similarly to https://github.com/frgfm/Holocron/blob/master/references/classification/README.md, if we specify on which dataset it was pretrained and with which image size, they can write that by themselves

test/test_inference.py
```diff
@@ -0,0 +1,34 @@
...
+from PIL import Image
+from pathlib import Path
+
+
+class PyronearPredictorTester(unittest.TestCase):
+
+    def setUp(self):
+        self.config = Path(__file__).parent.parent / 'pyrovision/inference/inference.cfg'
+        self.testImage = Path(__file__).parent / 'fixtures/wildfire_example.jpg'
+        print(self.config, self.testImage)
```
frgfm: You can add this at the top of the files in constants, no need to put it in the class attributes ;)

frgfm (commented):
Thanks for the PR!
I believe it would be easier to create a pyrovision/models/rexnet.py submodules where you write a `rexnet1_0x` function that instantiate the model properly and have a  `pretrained` argument similar to holocron or torchvision. What do you think?

We could move the preprocessing, post-processing + model loading part to pyroengine  :ok_hand:

pyrovision/inference/__init__.py
MateoLostanlen: Ok I get your point, I will make a rexnet module

pyrovision/inference/inference.cfg
MateoLostanlen: Of course you can't have one configuration file per model. The goal here is to do it only for our reference model. The one we use for all our predictions. This reference model will then be used wherever we need it on the raspberry pi, on the web app ...
What do you think about it ?

pyrovision/inference/pyronear_predictor.py
MateoLostanlen: As I said in the previous comment the goal is to create a unique predictor. This will only be changed when we get a better model.

test/test_inference.py
MateoLostanlen: ok ;)

pyrovision/inference/pyronear_predictor.py
MateoLostanlen: But I can remove the if statement, and just leave the transform I need

pyrovision/inference/inference.cfg
frgfm: Well my point is that this information could be stored as model attributes and set by constructor arguments. I'm not saying we don't need any of these. If I were to check each element here:
- backbone info : can be obtained by checking the class name
- num_classes : easily stored in the model or by checking the size of the last FC layer
- use_adaptive : as a constructor argument it's important but doesn't need to be stored afterwards (ie. we put this value as default in the function which will work with the pretrained model we have made)
- nb_features : is it to build the FC? if so, it can either be read as the last channel number in conv layers if there is a Global pooling, or dynamically by forwarding a mock input (and it is important because it depends then on the image size)
- cut : constructor /input arg is relevant, not sure it is useful as attribute
- url : should be stored in a static dict in the module
- device: i'd argue it should not be set. The model state_dict needs to be saved with CPU tensor by safety, but then the user will choose on which type of device she/he will use the model.
- transforms: input image size is relevant, but center crop isn't. Either you mean it as augmentation and it's for training, not inference, or we lose information in the image --> in which case this needs to be changed on inference: we don't want the model to say there is no fire if it's just outside the center crop while it was clearly in the input image and the model could have detected it.

All relevant information related to the model should be "linked" to it (as attributes). If we put them in a different folder/file, and forget to change them after updating a model, this will be troublesome :sweat_smile:  For me there are only two types of valuable information for inference purposes:
- how we build the model (constructor arguments, potentially used as attributes afterwards)
- how we prepare model's input / understand model outputs (image size, resizing/cropping scheme, normalization parameters, number of output classes)

The rest doesn't seem relevant for inference in my opinion :)
My apologies for the long reply haha

pyrovision/inference/pyronear_predictor.py
frgfm: cf. my previous reply
To sum up:
- inference information would be better stored as model attributes
- in pyronear-engine, we use the attributes (normalization parameters for instance) to create the pre-processing correctly and process outputs

Another possibility is to embed the preprocessing in the model, like in https://github.com/pytorch/vision/blob/master/torchvision/models/detection/faster_rcnn.py#L233 where there is a `transforms` module in the model that process the input tensor before passing it to the core model.

MateoLostanlen:
Hi @frgfm , I reorganized the PR. I propose you the following format, I added a pyronear_model module with a function to get the latest pyronear model. This way every time we have a new model we just have to modify this function only. I will move the pyronear_predictor in PyroEngine. What do you think ?

PS: I will correct the errors on the tests after, I just wanted to propose you this format.

pyrovision/models/pyronear_model.py
```diff
@@ -0,0 +1,34 @@
+ # -*- coding: utf-8 -*-
```
frgfm: My bad, I used to add those in the repo, but in Python3, this is no longer required

pyrovision/models/pyronear_model.py
```diff
@@ -0,0 +1,34 @@
...
+from pyrovision.models.utils import cnn_model
+import holocron
+from holocron.models.utils import load_pretrained_params
+
+
+## Define Model
+backbone = 'rexnet1_0x'
+num_classes = 1
+nb_features = 1280
+cut = -2
```
frgfm: All of those should be moved to default values of the function's argument :ok_hand:

pyrovision/models/pyronear_model.py
```diff
@@ -0,0 +1,34 @@
...
+url = "https://github.com/pyronear/pyro-vision/releases/download/v0.1.0/rexnet1_0x_acp_2e017f83.pth"
+
+
+def pyronear_model(pretrain=True, device='cpu'):
```
frgfm: pretrain --> pretrained would be more natural

pyrovision/models/pyronear_model.py
```diff
@@ -0,0 +1,34 @@
...
+    # Get backbone
+    base = holocron.models.__dict__[backbone](False, num_classes=num_classes)
+    # Change head
+    model = cnn_model(base, cut), nb_features=nb_features, num_classes=num_classes)
+    # Load Weight
+    if pretrain:
+        load_pretrained_params(model, url)
+    # Move to gpu
+    if device == 'cuda' and torch.cuda.is_available():
+        model = model.to('cuda:0')
```
frgfm: I think we should remove this completely. The user will be able to move the model to the correct device

test/test_models.py
```diff
@@ -110,6 +117,30 @@ def test_ssresnet18(self):
         self.assertEqual(out.shape[0], batch_size)
         self.assertEqual(out.shape[1], 1)
 
+    def test_pyronead_model(self):
+        # Define Model
+        model = models.pyronear_model()
+        model = model.eval()
+        sigmoid = nn.Sigmoid()
```
frgfm: you can remove this

test/test_models.py
```diff
@@ -110,6 +117,30 @@ def test_ssresnet18(self):
...
+                                     ])
+        # Load Image
+        im = Image.open(testImage)
+        im = tf(im).unsqueeze(0)
+
+        # Make Prediction
+        with torch.no_grad():
+            pred = model(im)
+
+        self.assertGreater(sigmoid(pred), 0.5)
```
frgfm: `self.assertGreater(torch.sigmoid(pred), 0.5)`

test/test_models.py
```diff
@@ -6,10 +6,17 @@
...
 import numpy as np
 import random
 from pyrovision import models
 from torchvision.models.resnet import BasicBlock
+from PIL import Image
+from pathlib import Path
+from torchvision import transforms
+
+# Define fire image to test models on a real use case
+testImage = Path(__file__).parent / 'fixtures/wildfire_example.jpg'
```
frgfm: This is specific to the test you added, I suggest moving it inside it if we don't use it for other unittests

frgfm (commented):
Thanks for the edits, this is starting to be very promising :)
A few minor adjustments and we're good to go!

pyrovision/models/pyronear_model.py
MateoLostanlen: There is this header everywhere in the repo, maybe we can remove this and addapt the header test in an other PR

pyrovision/models/pyronear_model.py
MateoLostanlen: ok 👍

pyrovision/models/pyronear_model.py
MateoLostanlen: ok

pyrovision/models/pyronear_model.py
frgfm: Forgot about the header check, I agree!

MateoLostanlen (commented):
@frgfm, I made the changes, thanks for your help ;)

pyrovision/models/__init__.py
```diff
@@ -8,3 +8,4 @@
 from .densenet import *
 from .mobilenet import *
 from .ssresnet import *
+from .pyronear_model import *
```
frgfm: Since this isn't an architecture definition, I think there are two options:
- either we name it model_zoo and we will add other pretrained models in here whatever their type
- or, since it's a rexnet, we create a module rexnet.py and you add a rexnet1_0 function and define it like in you pyronear_model

I think the latter is better considering we could trained different rexnet, and for a given rexnet type (let's say rexnet1_0), if we train a better one, it will replace the previous.

pyrovision/models/pyronear_model.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
```
frgfm: Let's rename the module to rexnet.py and name the function rexnet1_0x (pyronear_model is not informative)

pyrovision/models/pyronear_model.py
```diff
@@ -0,0 +1,33 @@
...
+# Copyright (c) Pyronear contributors.
+# This file is dual licensed under the terms of the CeCILL-2.1 and AGPLv3 licenses.
+# See the LICENSE file in the root of this repository for complete details.
+
+from pyrovision.models.utils import cnn_model
+import holocron
+from holocron.models.utils import load_pretrained_params
+
+
+url = "https://github.com/pyronear/pyro-vision/releases/download/v0.1.0/rexnet1_0x_acp_2e017f83.pth"
```
frgfm: I suggest replicating this: https://github.com/frgfm/Holocron/blob/master/holocron/models/rexnet.py#L13-L24

Here we create only a dict, with the function as a key (rexnet1_0x) and the url as a value. That way if we add another rexnet, we'll just have to add an entry.

pyrovision/models/pyronear_model.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def pyronear_model(backbone='rexnet1_0x', num_classes=1, nb_features=1280, cut=-2, pretrained=False):
+    r"""The purpose of this function is to return the latest Pyronear model.
+
+    Args:
+        backbone (string, optional): model backbone
```
frgfm: if we rename it to rexnet1_0x, you can remove the backbone argument

test/test_models.py
```diff
@@ -66,16 +70,18 @@ def _test_classification_model(self, name, input_shape):
...
-        # Default case
-        model = models.__dict__[name](num_classes=num_classes)
-        model.eval()
-        x = torch.rand(input_shape)
-        with torch.no_grad():
-            out = model(x)
-        # self.assertExpected(out, rtol=1e-2, atol=0.)
-        self.assertEqual(out.shape[-1], 50)
+        args = inspect.getfullargspec(models.__dict__[name])
+        if 'imagenet_pretrained' in args.args:
```
frgfm: This could be kept if passing imagenet_pretrained actually passes pretrained=True to holocron function

frgfm (changes requested):
Sorry about the delay, I didn't get the notification! There are a few minor adjustments to make, but the rest looks fine!

MateoLostanlen:
sorry @frgfm  i didn't saw your last review, i created the rexnet module, tell me if it's good for you

MateoLostanlen:
I have an error on the test_dataset but I didn't touch it in this PR, is it related to a recent modification ?

```
Traceback (most recent call last):
  File "/home/runner/work/pyro-vision/pyro-vision/test/test_datasets.py", line 87, in test_openfire
    self.assertAlmostEqual(len(train_set) + len(test_set), len(extract), delta=15)
AssertionError: 180 != 200 within 15 delta
```

pyrovision/models/rexnet.py
```diff
@@ -0,0 +1,153 @@
...
+                   'url': 'https://github.com/frgfm/Holocron/releases/download/v0.1.2/rexnet1_0x_224-ab7b9733.pth'},
+    'rexnet1_3x': {'width_mult': 1.3, 'depth_mult': 1.0,
+                   'url': 'https://github.com/frgfm/Holocron/releases/download/v0.1.2/rexnet1_3x_224-95479104.pth'},
+    'rexnet1_5x': {'width_mult': 1.5, 'depth_mult': 1.0,
+                   'url': 'https://github.com/frgfm/Holocron/releases/download/v0.1.2/rexnet1_5x_224-c42a16ac.pth'},
+    'rexnet2_0x': {'width_mult': 2.0, 'depth_mult': 1.0,
+                   'url': 'https://github.com/frgfm/Holocron/releases/download/v0.1.2/rexnet2_0x_224-c8802402.pth'},
+    'rexnet2_2x': {'width_mult': 2.2, 'depth_mult': 1.0,
+                   'url': None},
+}
```
frgfm: The architecture is unchanged vs. "rexnet1_0x" & the others in Holocron. It would be best to use the functions from there, the only diff is that we map pretrained differently:
- pretrained = True on our side should point to the checkpoint you added (thus passing pretrained=False to the Holocron rexnet functions)
- if pretrained=False, perhaps we can pass an additional args "imagenette_pretrained" which will pass its value to the "pretrained" of Holocron rexnet functions

pyrovision/models/rexnet.py
```diff
@@ -0,0 +1,153 @@
+# Copyright (C) 2021, Pyronear contributors.
+
+# This program is licensed under the GNU Affero General Public License version 3.
+# See LICENSE or go to <https://www.gnu.org/licenses/agpl-3.0.txt> for full license details.
+
+from holocron.models.rexnet import ReXNet
```
frgfm: You just need to import models, so that we can dynamically call the specific rexnet functions

pyrovision/models/rexnet.py
```diff
@@ -0,0 +1,153 @@
...
+
+    # Model creation
+    base_model = ReXNet(default_cfgs[arch]['width_mult'], default_cfgs[arch]['depth_mult'], **kwargs)
+
+    # Imagenet pretraining
+    if imagenet_pretrained:
+        if pretrained:
+            raise ValueError('imagenet_pretrained cannot be set to True if pretrained=True')
+
+        load_pretrained_params(base_model, default_cfgs[arch]['url'], progress)
```
frgfm: That section could be removed so that we pass the "imagenet_pretrained" value to the rexnet functions from Holocron

pyrovision/models/rexnet.py
```diff
@@ -0,0 +1,153 @@
...
+
+    # Parameter loading
+    if pretrained:
+        load_pretrained_params(model, model_urls[arch], progress=progress)
+
+    return model
+
+
+def rexnet1_0x(pretrained=False, progress=True, imagenet_pretrained=False, num_classes=1,
+               lin_features=512, dropout_prob=0.5, bn_final=False, concat_pool=True, **kwargs):
```
frgfm: starting from num_classes, the rest can be passed as **kwargs perhaps like in the other functions?

frgfm (commented):
Thanks for the edits! I added a few comments, we're almost there! Could you merge master as well please? the commit diff shows things that have been handled on master

pyrovision/models/rexnet.py
MateoLostanlen: yes makes sens :)

frgfm (approved):
Looks good thanks! We'll have to modify the unittest so that it also check the pretrained version when available, but we can do this later on!

PR #87, opened 2021-01-15
Update README.md

TekayaNidham:
wrong path


frgfm (approved):
Thanks for the PR! Good catch

Issue #94, opened 2021-01-27
[datasets] Module design suggestion


frgfm:
Here is a suggestion for the `pyrovision.datasets` module's organization:

- utils.py: shared utility functions
- XYZ.py: a VisionDataset inherited class definion with functions for dataset XYZ

This suggestion is based on the organization of https://github.com/pytorch/vision/tree/master/torchvision/datasets which is quite effective. The documentation should also include detailed explanation on how to use it.

If one dataset was made available by Pyronear, it should be accessible via a link with a hash for integrity check.
- [x] Refactors into a flat module (#97)
- [x] Assess which functions/classes should be kept in the repo (#136, #138)

TekayaNidham:
Good idea, that would work

![](https://media1.tenor.com/images/27768e286628639a13a74ebf7c51be8b/tenor.gif?itemid=13857370)

frgfm:
Hey @MateoLostanlen @blenzi @x0s @Akilditu ,

While refactoring the datasets module, I ended up wondering whether some things should be removed because of their misalignment with the repo's purpose. Generally speaking, the purpose of the datasets module in this repo is to make for each dataset its source accessible (using processed annotations if need be) and offering a torchvision dataset. I believe there are some features linked to wildfire that are not aligned with this (considering that for now, the source of wildfire is not even accessible publicly). Some details below:

**openfire**
- the URL needs to be moved to a release attachments

**video_utils**
- `FireLabeler`: for me, this helps to structure the dataset before exposing it publicly, but not after. In my opinion, this should be removed from the library.
- `FrameExtractor`: this is really helpful, but we could make it much better. First, do we want something to do this statically (creating image files from video files) or dynamically (used by the dataloader)? Second, whatever the first answer is, we need to make it compatible for other video datasets (perhaps there are interesting things we could use in https://github.com/pytorch/vision/blob/master/torchvision/datasets/video_utils.py). I would argue that considering its name, the parameters of the constructor should be sampling parameters (frames / sec, and max_frames for instance), and its call arguments should be a video file (or already read). What do you think?

**utils**
- with some regex, we could refactor or remove the name + extension resolution

**wildfire**
[...]

PR #96, opened 2021-01-31
chore: Moved environment collection script outside of lib

frgfm:
This PR introduces the following modifications:
- moved collect_env from `pyrovision.utils` to the `scripts` folder
- fixed URL in bug report template
- refactored collect_env script for leaner information display

Successfully merging this PR will close #91 

Any feedback is welcome!

PR #97, opened 2021-02-03
refactor: Refactored pyrovision.datasets module

frgfm:
Following up on #94, this PR introduces the following modifications:
- refactors the datasets module into: utils (general utilities functions), video_utils (utilities specific to videos), openfire (everything related to OpenFireDataset), wildfire (everything related to WildFireDataset)
- reflected these changes in unittests

Once this PR is merged (which keeps every single class/function), we'll need to reassess which part are relevant to the library I think (fire labeler might not be really relevant here, perhaps in a repo dedicated to data labelling?)
Any feedback is welcome!


frgfm:
The link validity check failure isn't related, and will be fixed once #96 is merged. Regarding the unittest failure, I have seen this one of other PRs before, I'm not sure about its origin. It's seems like a random edge case because it doesn't always pop, and I'm not able to reproduce it locally.

MateoLostanlen (approved):
Hi, fg and thank you for the PR. It seems likely that future datasets will have the same metadata style as wilfire, so we can easily reuse the fire labeleller with some minor adaptations. Therefore, this update seems relevant to me.

All good for me!

PR #98, opened 2021-02-07
fix: Fixed CUDA collection for v11+

frgfm:
As mentioned @MateoLostanlen in #96, the CUDA version collection was failing with our current version of the script. As suspected, this is due to the regex used not being compatible with nvcc console output after version 11 cf. https://github.com/pytorch/pytorch/issues/51840

This PR fixes this using what was mentioned in the above issue, and now works both for CUDA>= 11 and CUDA < 11

PR #100, opened 2021-02-07
test: Change sample download failure tolerance for OpenFire

### LINKED ISSUE #99: To many dead links in OpenFire Dataset
## 🐛 Bug

As we can see in PR #98 and #86 currently the test_openfire test does not pass anymore. The number of dead links is 17 which exceeds the limit of 15 set in the test

MateoLostanlen:
I propose here a quick repair for issue #99 , it's obviously not a stable solution but it allows us to move forward on the rest. It's not a top priority because we don't use openfire anymore but we should download the dataset and put it in a safe place to see that this problem is getting worse.


frgfm (approved):
I agree, I intended to open the same PR! Thanks :ok_hand: The last test failure will be fixed once #98 is merged

PR #102, opened 2021-02-10
refactor: Replaced Flatten by torch version

frgfm:
This PR simply replaces the in-house Flatten module implementation by torch one, now that one has been added to the framework.


MateoLostanlen (approved):
👍

PR #104, opened 2021-02-13
refactor: Reorganized references folder

### LINKED ISSUE #103: [reference] Reorganize reference folder
@frgfm, regarding your comment:
https://github.com/pyronear/pyro-vision/pull/101#pullrequestreview-588353780

Maybe we should reorganize the reference folder too, we can split it by tast (classification, detection ...) as it is done in holocron

Then we can specify the dataset in a leaderbord

MateoLostanlen:
The purpose of this PR is to close #103


frgfm (commented):
Thanks for the PR! While I agree the priority should be set to WildFire, considering the dataset version we use for training is not yet available publicly in the lib, I suggest keeping the OpenFire training by merging the two scripts (having an option pointing to the dataset the user wants to train on, wildfire/lightfire by default). What do you think?

MateoLostanlen:
ok I understand your point I'll make the change

MateoLostanlen:
Hi fg, I added the possibility to train a model with openfire. We need to make Wildfire downloadable in the same way (with a flag download=True) in a futur PR.

MateoLostanlen:
There is an issue with doc build, there is already an open issue on the repo : https://github.com/sphinx-doc/sphinx/issues/8885

references/classification/train.py
```diff
@@ -124,6 +133,7 @@ def parse_args():
                                      formatter_class=argparse.ArgumentDefaultsHelpFormatter)
 
     parser.add_argument('data_path', type=str, help='path to dataset folder')
+    parser.add_argument('use_OpenFire', type=bool, default=False, help='use our open source OpenFire dataset')
```
frgfm: always use lower case in arguments name :)
and here you would need to do `--use_OpenFire True` for this to work

I'd suggest
```
parser.add_argument('use-openfire', dest="use_openfire", action="store_true", help='use our open source OpenFire dataset')
```

so that we only have to do `--use-openfire`

references/classification/README.md
```diff
@@ -17,14 +17,23 @@ Install pyro-vision first
...
+```
 
-## Train
+### OpenFire
+
+You can also use out opensource dataset without password
 
 ```
-python train.py WildFireLght/ --model rexnet1_0x --lr 1e-3 -b 16 --epochs 20 --opt radam --sched onecycle --device 0
+python train.py OpenFire/ --use_OpenFire True --model rexnet1_0x --lr 1e-3 -b 16 --epochs 20 --opt radam --sched onecycle --device 0
```
frgfm: This will have to be edited with the change of args, the rest looks good :+1:

frgfm (changes requested):
Small last change and we're good to go!

frgfm:
Watch out not to mix PR purposes together, I just noticed that you removed utils.rst in the documentation. This should be done in a separate PR as this is a completely different topic, I'll open one to handle this quickly :ok_hand: 

Best practices:
- 1PR = 1 goal
- the goal should be put in the title, and the description should explain exhaustively all modifications

Otherwise it gets very hard quickly to track back past modifications :/

references/classification/train.py
```diff
@@ -28,7 +30,7 @@ def target_transform(target):
     return target.unsqueeze(dim=0)
 
 
-def load_data(traindir, valdir, img_size=224, crop_pct=0.8):
+def load_data(data_path, use_OpenFire=False, download=False, img_size=224, crop_pct=0.8):
```
frgfm: You can remove `download=False` and perhaps replace `user_openfire` all lowercase :ok_hand:

references/classification/train.py
```diff
@@ -48,8 +50,17 @@ def load_data(traindir, valdir, img_size=224, crop_pct=0.8):
                                          ])
 
     print("Loading data")
-    train_set = ImageFolder(traindir, train_transforms, target_transform=target_transform)
-    val_set = ImageFolder(valdir, val_transforms, target_transform=target_transform)
+    if use_OpenFire:
```
frgfm: to be replaced

frgfm (commented):
All good, I think you merged unwanted modifications as well (cf. comments). But it's really nitpicking

references/classification/train.py
```diff
@@ -68,8 +68,11 @@ def load_data(data_path, use_OpenFire=False, download=False, img_size=224, crop_
 def main(args):
 
     print(args)
+    if args.use-openfire:
+        use-openfire=True
```
frgfm: you can remove this

references/classification/train.py
```diff
@@ -68,8 +68,11 @@ def load_data(data_path, use_OpenFire=False, download=False, img_size=224, crop_
...
 
-    train_set, val_set = load_data(args.data_path, use_OpenFire=args.use_OpenFire, img_size=args.img_size)
+
+    train_set, val_set = load_data(args.data_path, use_openfire=use-openfire, img_size=args.img_size)
```
frgfm: simply pass `args.use_openfire`, no need to complicate this :sweat_smile:

frgfm (changes requested):
Sorry my previous feedback wasn't clear apparently :sweat_smile:

frgfm:
Also, I didn't realize, but you have to update the instructions of the "References" section in the main README ;)

references/classification/train.py
MateoLostanlen: WTF?? no idea why I did this ... haha

README.md
```diff
@@ -62,32 +62,41 @@ pip install pyronear
 
 ### datasets
 
-Access all pyrovision datasets just like any `torchvision.datasets.VisionDataset`:
+Access out OpenFire dataset just like any `torchvision.datasets.VisionDataset`:
```
MateoLostanlen: not all for now, we need to do the same for the wildfire dataset !

frgfm: Indeed, just a little typo "Access out OpenFire..." -> "Access the OpenFire ..."

frgfm (commented):
One last typo, and we can merge :sweat_smile:

README.md
frgfm: Actually, perhaps don't modify this, the README was edited in another ongoing PR

frgfm (dismissed):
Looks good to me! Seems like you'll need to merge master before moving on though

frgfm (approved):
Thanks for the edits!

PR #106, opened 2021-02-15
docs: Fixed sphinx requirements

frgfm:
This PR simply adds a version constraint on sphinx since the latest release is causing issues. The bug has already been reported but won't be fixes until next release, hence this new constraint.


MateoLostanlen (approved):
Thanks for the fix !

PR #108, opened 2021-02-17
feat: Added Dockerfile for ubuntu image

frgfm:
- adds a Dockerfile for a basic ubuntu image
- updated README
- added CI job to check that the docker builds successfully

Any feedback is welcome!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -110,3 +110,13 @@ jobs:
       - name: Build documentation
         run: |
           sphinx-build docs/source docs/build -a
+
+  docker-ready:
+    runs-on: ubuntu-latest
+    needs: pkg-install
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build docker image
+        run: docker build . -t pyrovision-py3.8.1-torch1.7-slim
+      - name: Run docker container
+        run: docker run pyrovision-py3.8.1-torch1.7-slim python -c 'import pyrovision'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,4 +119,4 @@ jobs:
       - name: Build docker image
         run: docker build . -t pyrovision-py3.8.1-torch1.7-slim
       - name: Run docker container
-        run: docker run pyrovision-py3.8.1-torch1.7-slim python -c 'import pyrovision'
+        run: docker run pyrovision-py3.8.1-torch1.7-slim python -c "import pyrovision; print(pyrovision.__version__)"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM python:3.8.1-slim
+
+ENV PYTHONUNBUFFERED 1
+ENV PYTHONDONTWRITEBYTECODE 1
+
+
+COPY ./requirements.txt /tmp/requirements.txt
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM python:3.8.1-slim
+
+ENV PYTHONUNBUFFERED 1
+ENV PYTHONDONTWRITEBYTECODE 1
```
chloeskt, line 4: Thank you @frgfm for this PR ! 

Small question: what is the use of this env variable in our case ?

frgfm: It prevents the generation of `.pyc` upon compilation :)

chloeskt: Yes, but why should we add this ? only to save disk space ?

frgfm: yes, and simply to keep the number of files to a minimum. It's a very common thing to add in a dockerfile

chloeskt: Ok thanks ! 😄

chloeskt (dismissed):

README.md
```diff
@@ -60,9 +60,9 @@ pip install pyronear
 
 ## Usage
 
-### datasets
+### Python package
 
-Access all pyrovision datasets just like any `torchvision.datasets.VisionDataset`:
+You can use the library like any other python package to analyze your documents as follows:
```
MateoLostanlen, line 65: Why did you removed this here ?

.github/workflows/main.yml
```diff
@@ -110,3 +110,13 @@ jobs:
...
+
+  docker-ready:
+    runs-on: ubuntu-latest
+    needs: pkg-install
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build docker image
+        run: docker build . -t pyrovision-py3.8.1-torch1.7-slim
+      - name: Run docker container
+        run: docker run pyrovision-py3.8.1-torch1.7-slim python -c 'import pyrovision'
```
MateoLostanlen: 'import pyrovision' gives an error on my machine (SyntaxError: EOL while scanning string literal), I need to use "import pyrovision" instead

.github/workflows/main.yml
```diff
@@ -110,3 +110,13 @@ jobs:
...
         run: |
           sphinx-build docs/source docs/build -a
+
+  docker-ready:
+    runs-on: ubuntu-latest
+    needs: pkg-install
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build docker image
+        run: docker build . -t pyrovision-py3.8.1-torch1.7-slim
```
MateoLostanlen, line 120: The bild work well, thanks for that :)

README.md
frgfm: I was starting to think that this is the purpose of the documentation, not the README. Usage instructions should be plain user-friendly examples of how to use the package in my opinion.

.github/workflows/main.yml
frgfm: Oh good to know, it means that this behaves differently on windows then! I'll fix this right away

README.md
MateoLostanlen: ok ok

MateoLostanlen (approved):

MateoLostanlen:
All good :)

PR #114, opened 2021-03-25
update requierements

MateoLostanlen:
If we open a collab notebook and install pyrovision

`pip install pyrovision`

Then try to import mobilenet_v2 `from pyrovision.models.mobilenet import mobilenet_v2 ` ,  that leed to the following error:

```
 ImportError: cannot import name 'model_urls' from 'torchvision.models.mobilenet' (/usr/local/lib/python3.7/dist-packages/torchvision/models/mobilenet.py)
```

We need to update torchvision in the requirements to fix it

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 opencv-python>=3.4.5.20
 pandas>=0.25.2
-torch>=1.2.0
-torchvision>=0.4.0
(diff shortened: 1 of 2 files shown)


MateoLostanlen:
Bad interpretation on my part the error does not come from there, it is a change of import

PR #117, opened 2021-03-28
read setup requirements in requirements.txt

MateoLostanlen:
Let's avoid useless code duplication that can lead to errors :)


frgfm:
@MateoLostanlen actually the duplication is on purpose since this is the only way to build the conda distrib properly :sweat_smile:

MateoLostanlen:
good to know :)

PR #119, opened 2021-03-28
increase OpenFire download threshold

MateoLostanlen:
As explain in : https://github.com/pyronear/pyro-vision/issues/118#issue-842791057


frgfm (approved):
Thanks!

PR #121, opened 2021-04-15
open_firev2

### LINKED ISSUE #9: [ci] Optimize cache for datasets downloading
## 🚀 Feature
Save downloaded datasets in CI cache.

## Motivation

With the addition of `datasets.OpenFire`, the test require the download of the entire dataset. Considering that the samples are scattered in multiple URLs, the download is taking at least several minutes.

## Pitch

- [ ] find a unique identifier to checksum all dataset URLs
- [ ] save cache once the job is completed to avoid downloading again if the file is the same

### LINKED ISSUE #118: Issue downloading OpenFire
## 🐛 Bug

This test do not pass anymore : 

We have a problem with OpenFire, the links to the images work less and less.

Now this test doesn't work anymore:
https://github.com/pyronear/pyro-vision/blob/db08df721c0e563117bded71c71b87b3f1d23b3a/test/test_datasets.py#L103

In the past we have already increased the threshold from 15 to 20 to get around this problem.

I propose to do it again with a big margin, let's take 30.

On the other hand we have to absolutely store this dataset somewhere soon and change the download mode.

MateoLostanlen:
Open_fire v1 is dying due to too many dead links as discuss in issue #118. I want to propose here a Open_fire V2. I have a dataset of 9000 images from which a good part comes from openfire. I propose that this dataset becomes the new Open_fire. It is downloadable from my drive so no problem with dead link here.

I guess this PR should close #9 too.

PS: For testing purposes I have created a sample version that can be downloaded quickly.
PPS: The real version of the datatset is not yet on my drive I have to put it in the next coming days but I have wifi problems

After this PR I will use the sample template for Wilfire dataset


frgfm:
Thanks for the PR and the dataset!
Though, GDrive is not exactly the best solution for dataset storage. How big is it? 
I would rather consider putting it in a release attachment and add integrity check :ok_hand:

mateoIdemia:
@frgfm 500Mb, it is too big ?

MateoLostanlen:
@frgfm 500Mb, it seems too big : https://docs.github.com/en/github/managing-your-work-on-github/file-attachments-on-issues-and-pull-requests

frgfm:
@MateoLostanlen 500Mb is quite alright :) 
- first, the limitation you are mentioning is about attachment to a comment/PR/issues. I'm positive the limitation for releases are higher since I have model checkpoints of my own bigger than 150Mb
- worst case scenario if it's a bit too big, we can easily split it in 2/3 parts and put each part in the release attachments !

MateoLostanlen:
@frgfm ok nice !
Should we add it to this repo or the dataset [repo](https://github.com/pyronear/pyro-datasets-builder) we are creating ?

MateoLostanlen:
Hi @frgfm, what do you thinks of these last changes ?

pyrovision/datasets/__init__.py
```diff
@@ -1,3 +1,2 @@
 from .openfire import OpenFire
 from .wildfire import WildFireDataset
-from . import utils
```
frgfm, line 2: mmh if we still have a utils submodule, I think we need to keep this

test/test_datasets_utils.py
```diff
@@ -1,53 +0,0 @@
-# Copyright (C) 2021, Pyronear contributors.
```
frgfm: we still have a utils submodule if we merge this PR, so some unittests would be good :)

pyrovision/datasets/utils.py
```diff
@@ -3,153 +3,34 @@
...
+    with DownloadProgressBar(unit='B', unit_scale=True,
+                             miniters=1, desc=url.split('/')[-1]) as t:
+        urllib.request.urlretrieve(url, filename=output_path, reporthook=t.update_to)
+
+        # check integrity of downloaded file
+        with open(output_path, 'rb') as f:
+            sha_hash = hashlib.sha256(f.read()).hexdigest()
+        if sha_hash[:8] != url.split('-')[-1].split('.')[0]:
+            raise RuntimeError("File not found or corrupted.")
+        print("\nUnziping ...")
```
frgfm: We could potentially add console logging, but let's avoid printing :sweat_smile:

pyrovision/datasets/utils.py
```diff
@@ -3,153 +3,34 @@
...
+        with zipfile.ZipFile(output_path, 'r') as zip_ref:
+            zip_ref.extractall(root)
+        print('Done!')
```
frgfm: same here

pyrovision/datasets/openfire.py
```diff
@@ -4,195 +4,69 @@
...
-                          'Please retry downloading later.')
-        # Extract errors
-        if target_issues > 0:
-            warnings.warn(f'{target_issues}/{len(extract)} samples have corrupted targets.')
-
-        return valid_samples
+        if self.train:
+            self.path_to_imgs = Path(self.root, 'train')
+        else:
+            self.path_to_imgs = Path(self.root, 'test')
```
frgfm:
```
self.path_to_imgs = Path(self.root, 'train' if self.train else 'test')
```

pyrovision/datasets/openfire.py
```diff
@@ -4,195 +4,69 @@
...
 
+def pil_loader(path: str) -> Image.Image:
+    # open path as file to avoid ResourceWarning (https://github.com/python-pillow/Pillow/issues/835)
+    with open(path, 'rb') as f:
+        img = Image.open(f)
+        return img.convert('RGB')
 
-class OpenFire(VisionDataset):
+
+class OpenFire(DatasetFolder):
```
frgfm: In the process of updating the class, several methods were removed. Are they covered by DatasetFolder? (len, getitem, etc.)

pyrovision/datasets/utils.py
```diff
@@ -3,153 +3,34 @@
...
-        entries (list<str, str>): URL and destination filen
-        root (pathlib.Path): folder where the files will be saved in
-        timeout (float, optional): number of seconds before the request times out
-        retries (int, optional): number of additional allowed download attempts
-        threads (int, optional): number of threads to be used for multiprocessing
-        silent (bool, optional): whether Exception should be raised upon download failure
-    """
-
-    parallel(partial(download_url, root=root, timeout=timeout, retries=retries, silent=silent),
-             entries, threads=threads)
```
frgfm: Even if OpenFire doesn't use this anymore, this could still be quite useful in the future (downloading list of files in parallel is much faster than doing this sequentially), and I don't see many reasons to remove this :sweat_smile:

pyrovision/datasets/utils.py
```diff
@@ -3,153 +3,34 @@
...
+    output_path = url.split('/')[-1]
+    with DownloadProgressBar(unit='B', unit_scale=True,
+                             miniters=1, desc=url.split('/')[-1]) as t:
+        urllib.request.urlretrieve(url, filename=output_path, reporthook=t.update_to)
+
+        # check integrity of downloaded file
+        with open(output_path, 'rb') as f:
+            sha_hash = hashlib.sha256(f.read()).hexdigest()
+        if sha_hash[:8] != url.split('-')[-1].split('.')[0]:
+            raise RuntimeError("File not found or corrupted.")
```
frgfm: Downloading & extracting are two independent things: I suggest updating the previous function for download, and if you need to compose them, adding a function calling the latter :ok_hand:

frgfm: Let's leverage existing functions from torchvision if possible : https://github.com/pytorch/vision/blob/master/torchvision/datasets/utils.py

test/test_datasets.py
```diff
@@ -68,51 +66,15 @@ def get_wildfire_image():
...
-            self.assertIsInstance(train_set, VisionDataset)
-            # Check unicity of sample across all splits
-            train_paths = [sample['name'] for sample in train_set.data]
-            self.assertTrue(all(sample['name'] not in train_paths for sample in test_set.data))
+            ds = datasets.OpenFire(root=root, download=True, train=True, sample=True, img_folder=None)
+            self.assertEqual(len(ds), 100)
+            ds = datasets.OpenFire(root=root, download=False, train=False, sample=True, img_folder=None)
+            self.assertEqual(len(ds), 16)
+            ds = datasets.OpenFire(root=root, download=False, train=False, sample=True, img_folder=root)
+            self.assertEqual(len(ds), 16)
```
frgfm: This is a good starting point for a unittest, but we're only checking that the constructor works here. We need to verify that its expected usage (retrieving a sample by indexing for instance) is correctly done. I suggest keeping the inspiration from the previous unittest and adapting it :ok_hand:

pyrovision/datasets/openfire.py
```diff
@@ -4,195 +4,69 @@
...
-from torchvision.datasets import VisionDataset
-from .utils import download_url, download_urls
+__all__ = ['OpenFire']
 
-ImageFile.LOAD_TRUNCATED_IMAGES = True
 
-__all__ = ['OpenFire']
+url = "https://github.com/pyronear/pyro-vision/releases/download/v0.1.2/open_fire-dd0b5bb0.zip"
+url_sample = "https://github.com/pyronear/pyro-vision/releases/download/v0.1.2/open_fire_sample-06a07e48.zip"
+zip_filename = "open_fire.zip"
```
frgfm: Let's move those as class attributes of OpenFire :+1:

pyrovision/datasets/openfire.py
```diff
@@ -4,195 +4,69 @@
...
+
 
+def pil_loader(path: str) -> Image.Image:
+    # open path as file to avoid ResourceWarning (https://github.com/python-pillow/Pillow/issues/835)
+    with open(path, 'rb') as f:
+        img = Image.open(f)
```
frgfm: let's add a `mode='r'` as argument to avoid issues later on

pyrovision/datasets/openfire.py
```diff
@@ -4,195 +4,69 @@
...
 
-ImageFile.LOAD_TRUNCATED_IMAGES = True
 
-__all__ = ['OpenFire']
+url = "https://github.com/pyronear/pyro-vision/releases/download/v0.1.2/open_fire-dd0b5bb0.zip"
+url_sample = "https://github.com/pyronear/pyro-vision/releases/download/v0.1.2/open_fire_sample-06a07e48.zip"
+zip_filename = "open_fire.zip"
+
 
+def pil_loader(path: str) -> Image.Image:
```
frgfm: `path: Union[str, Path]` I believe

pyrovision/datasets/openfire.py
```diff
@@ -4,195 +4,69 @@
...
+            self.path_to_imgs = Path(self.root, 'train')
+        else:
+            self.path_to_imgs = Path(self.root, 'test')
 
-    def extra_repr(self):
-        return "Split: {}".format("Train" if self.train is True else "Test")
+        super(OpenFire, self).__init__(self.path_to_imgs, loader, None,
+                                       transform=transform,
+                                       target_transform=target_transform,
+                                       is_valid_file=loader)
```
frgfm: I'm not really familiar with DatasetFolder. Do you have any ref of implemented dataset of torchvision using it so I can dive into this better please ? :pray:

frgfm (commented):
Thanks for the PR! I have many questions as I'm not familiar with some classes that you used in here :sweat_smile:

pyrovision/datasets/openfire.py
MateoLostanlen: I guess you know [ImageFolder](https://github.com/pytorch/vision/blob/4ec38d496db69833eb0a6f144ebbd6f751cd3912/torchvision/datasets/folder.py#L177), which is [recommended by god](https://discuss.pytorch.org/t/custom-label-for-torchvision-imagefolder-class/52300/2) by the way :) The goal is to adapt the ImageFolder construction to our problem.

pyrovision/datasets/openfire.py
MateoLostanlen: Why? So the user of our dataset has to fill in a link that will never change

test/test_datasets.py
MateoLostanlen: Ok, I'm on it :)

pyrovision/datasets/utils.py
MateoLostanlen: It was not used anymore, but I can leave it if you want :)  Just I need to setup a "fake test" for it since it's not tested in OpenFire anymore

pyrovision/datasets/openfire.py
MateoLostanlen: yep : https://github.com/pytorch/vision/blob/730c5e1eab130e2900c8e839ea08fa11f024516f/torchvision/datasets/folder.py#L223

pyrovision/datasets/openfire.py
frgfm: I don't mean class constructor arguments, I do mean fixed class attributes :)

frgfm: In case we end up with different versions of OpenFire, there will be URLs for each of them, so it makes more sense to have those as class attributes than module variables I think

MateoLostanlen:
Hi @frgfm , I changed everything to use the torchvision dataset template. Tell me if it's ok for you

I'm thinking of doing the same thing in another PR for wildfire with the wildfire class that inherits from openfire. Like it is done for [MNIST and FashionMNIST](https://github.com/pytorch/vision/blob/26e34735bc935b0d54588dc32358171390ab593b/torchvision/datasets/mnist.py#L196). By adding the right urls and the possibility to pass a password to decrypt the zip

test/test_datasets.py
```diff
@@ -68,51 +66,13 @@ def get_wildfire_image():
...
-            self.assertIsInstance(img, Image)
-            self.assertIsInstance(target, int)
-            self.assertEqual(train_set.class_to_idx[extract[0]['target']], target)
-
-            # Check train/test split
-            self.assertIsInstance(train_set, VisionDataset)
-            # Check unicity of sample across all splits
-            train_paths = [sample['name'] for sample in train_set.data]
-            self.assertTrue(all(sample['name'] not in train_paths for sample in test_set.data))
+            print(root)
```
frgfm: Let's avoid any console print

pyrovision/datasets/openfire.py
```diff
@@ -20,179 +19,125 @@ class OpenFire(VisionDataset):
...
-                          'Please retry downloading later.')
-        # Extract errors
-        if target_issues > 0:
-            warnings.warn(f'{target_issues}/{len(extract)} samples have corrupted targets.')
-
-        return valid_samples
-
-    def extra_repr(self):
+    def extra_repr(self) -> str:
         return "Split: {}".format("Train" if self.train is True else "Test")
```
frgfm: Let's use modern Python3.6 syntax and avoid uppercase for variables
```python
return f"train={self.train}"
```

which will render this `print(dataset)` as
```shell
OpenFire(train=True)
```

pyrovision/datasets/openfire.py
```diff
@@ -20,179 +19,125 @@ class OpenFire(VisionDataset):
...
-            valid_samples (list<dict>): list of valid samples
-        """
+        try:
+            download_and_extract_archive(
+                self.url, download_root=self.raw_folder,
+                filename=self.filename,
+                md5=md5
+            )
+        except URLError as error:
+            print(
```
frgfm: no print, but we can use `logging.warn`

frgfm: And I'm not sure letting this error splip through would be a good idea: if the user isn't able to download & verify the dataset, that person might start a training on corrupt data. Perhaps letting the error go off would be better

pyrovision/datasets/openfire.py
```diff
@@ -20,179 +19,125 @@ class OpenFire(VisionDataset):
...
+                "Failed to download (trying next):\n{}".format(error)
```
frgfm: If we still catch the error, let's use modern python syntax is faster & better: 
```python
f"failed to download {self.url}. Please investigate"
```

pyrovision/datasets/openfire.py
```diff
@@ -20,179 +19,125 @@ class OpenFire(VisionDataset):
...
+        else:
+            self.url = self.urls['ds']
+            md5 = self.md5s['ds']
 
-        Args:
-            extract (list<dict>): image extract to download
-            threads (int, optional): number of threads used for parallel downloading
-        """
+        path = os.path.normpath(self.url)
+        self.filename = path.split(os.sep)[-1]
```
frgfm: faster alternative
```python
self.filename = path.rpartition(os.sep)[-1]
```

pyrovision/datasets/openfire.py
```diff
@@ -20,179 +19,125 @@ class OpenFire(VisionDataset):
...
+
+    def download(self) -> None:
+        """Download the OpenFire data if it doesn't exist already."""
+        # download files
+        if self.sample:
+            self.url = self.urls['sample']
+            md5 = self.md5s['sample']
+        else:
+            self.url = self.urls['ds']
+            md5 = self.md5s['ds']
```
frgfm:
```python
self.url = self.urls['sample' if self.sample else 'ds']
md5 = self.md5s['sample' if self.sample else 'ds']
```

pyrovision/datasets/openfire.py
```diff
@@ -20,179 +19,125 @@ class OpenFire(VisionDataset):
...
 
         if download:
-            self.download(threads, num_samples)
+            self.download()
 
-        # Load appropriate subset
-        extract = [sample for sample in self.get_extract(num_samples)
-                   if sample['is_test'] == (not train)]
+        if not self._check_exists():
+            raise RuntimeError('Dataset not found.' +
```
frgfm: `FileNotFoundError` would be a better error type :+1:

pyrovision/datasets/openfire.py
```diff
@@ -20,179 +19,125 @@ class OpenFire(VisionDataset):
...
-            self.img_folder = Path(img_folder)
+    urls = {'ds': 'https://github.com/pyronear/pyro-vision/releases/download/v0.1.2/open_fire.zip',
+            'sample': 'https://github.com/pyronear/pyro-vision/releases/download/v0.1.2/open_fire_sample.zip'
+            }
+
+    md5s = {'ds': '5a532853ac17dc43ed7dd4a97a15d715',
+            'sample': '31117230cceb029d557a1981f0f30cf7'
+            }
+
+    classes = ['0 - No Fire', '1 - Fire']
```
frgfm: The index will be outputed, so I don't think we need the figure here
```python
classes: List[str] = ['No fire', 'Fire']
```

pyrovision/datasets/openfire.py
```diff
@@ -20,179 +19,125 @@ class OpenFire(VisionDataset):
...
+
+    def __init__(
+            self,
+            root: str,
+            train: bool = True,
```
frgfm: There is no difference whatever the value of `train`. Perhaps we should remove this

pyrovision/datasets/openfire.py
```diff
@@ -20,179 +19,125 @@ class OpenFire(VisionDataset):
...
-            self.img_folder = Path(self.root, self.__class__.__name__, 'images')
-        else:
-            self.img_folder = Path(img_folder)
+    urls = {'ds': 'https://github.com/pyronear/pyro-vision/releases/download/v0.1.2/open_fire.zip',
+            'sample': 'https://github.com/pyronear/pyro-vision/releases/download/v0.1.2/open_fire_sample.zip'
+            }
+
+    md5s = {'ds': '5a532853ac17dc43ed7dd4a97a15d715',
+            'sample': '31117230cceb029d557a1981f0f30cf7'
+            }
```
frgfm: I would rather suggest:

```python
DATA = ("https://github.com/pyronear/pyro-vision/releases/download/v0.1.2/open_fire.zip", "5a532853ac17dc43ed7dd4a97a15d715")
SAMPLE = (''https://github.com/pyronear/pyro-vision/releases/download/v0.1.2/open_fire_sample.zip'', "31117230cceb029d557a1981f0f30cf7")
```

and updating the rest accordingly

frgfm (commented):
Thanks for the edits! I added extra comments, sorry I has to take some time diving back in :sweat_smile:

pyrovision/datasets/openfire.py
MateoLostanlen: there is a difference [here](https://github.com/MateoLostanlen/pyro-vision/blob/2a71cc6a7e2ace51f7ffa3d472d91466fa9d2973/pyrovision/datasets/openfire.py#L72)

pyrovision/datasets/openfire.py
MateoLostanlen: Ok I get your point, I have removed the try catch

pyrovision/datasets/openfire.py
MateoLostanlen: hello @frgfm I guess it's time to end this PR :) what is the point of using a tuple instead of a dictionary?

frgfm: If you miss an entry out of 2 in my proposition: the data will be working perfectly, the sample will crash right away (no entry available)

with both dict, if you miss an entry in one of the dict (or mess up one of the dict), nothing will work anymore 

See it as a safeguard :) but nothing critical, we can ignore that for now if it's the only blocking aspect

MateoLostanlen: No just wanted to understand your logic, I get it I will make the change. Thanks :)

frgfm:
@MateoLostanlen sorry I thought that you saw, but I didn't review because there is still a failing test which is related to this PR. Would you mind fixing it before moving on to the review? :pray:

MateoLostanlen:
@frgfm  oh sorry this pr is open for too long I had forgotten :) It's ok the error is fixed. There is still a problem but it has nothing to do with pyrovision as explained [here](https://stackoverflow.com/questions/70344739/backend-youtube-dl-py-line-54-in-fetch-basic-self-dislikes-self-ydl-info)

test/test_datasets.py
```diff
@@ -68,51 +66,13 @@ def get_wildfire_image():
...
-            # Check train/test split
-            self.assertIsInstance(train_set, VisionDataset)
-            # Check unicity of sample across all splits
-            train_paths = [sample['name'] for sample in train_set.data]
-            self.assertTrue(all(sample['name'] not in train_paths for sample in test_set.data))
+
+            ds = datasets.OpenFire(root=root, train=True, transform=tf, download=True, sample=True)
+            self.assertEqual(len(ds), 64)
+            x, _ = ds[0]
+            self.assertEqual(x.shape[1], 224)
```
frgfm: Out of safety, I think we should add additional test for both the image & the target: checking whether they are tensors, checking their data type at least :sweat_smile: (considering we've dropped a LOT of other checks haha)

pyrovision/datasets/openfire.py
```diff
@@ -20,179 +18,114 @@ class OpenFire(VisionDataset):
...
-    def __init__(self, root, train=True, download=False, threads=None, num_samples=None,
-                 img_folder=None, **kwargs):
-        super(OpenFire, self).__init__(root, **kwargs)
-        self.train = train
-        if img_folder is None:
-            self.img_folder = Path(self.root, self.__class__.__name__, 'images')
-        else:
-            self.img_folder = Path(img_folder)
+    DATA = ("https://github.com/pyronear/pyro-vision/releases/download/v0.1.2/open_fire.zip",
+            "5a532853ac17dc43ed7dd4a97a15d715")
```
frgfm: I just inspected the images, and I'm a bit concerned whether we can distribute those. Some come from Shutterstock for instance

Previously we were only sharing the actual URL, but here we share the actual content and I'm not sure we can :/

frgfm (commented):
Thanks! My bad, I had to jump back into this

pyrovision/datasets/openfire.py
MateoLostanlen: I don't know if it make a difference .. @fe51 any legal advice ?

test/test_datasets.py
MateoLostanlen: I have added some tests tell me what you think :)

pyrovision/datasets/openfire.py
MateoLostanlen: If this is not possible we will have to make the dataset private too. The solution with urls is not stable enough in all cases

frgfm (commented):
The unittests looks good thanks!

For now, about the legal part of sharing the URL of the dataset, I can suggest a viable option : we don't share the URL, we only specify about the folder should be structured. People make their own dataset and they'll have to pass the path to the extracted archive.

(while we figure out how/if we can share this)

MateoLostanlen:
I thinks it will be complicated [in a legal point of view ](https://qr.ae/pGLYAj). I think our current version it's not legal either. So let's delete it from the repo. However if you do not download any images, this dataset became exactly the [ImageFolder](https://pytorch.org/vision/main/generated/torchvision.datasets.ImageFolder.html) dataset from torchvision. It is still usefull ?

frgfm:
> I thinks it will be complicated [in a legal point of view ](https://qr.ae/pGLYAj). I think our current version it's not legal either. So let's delete it from the repo. However if you do not download any images, this dataset became exactly the [ImageFolder](https://pytorch.org/vision/main/generated/torchvision.datasets.ImageFolder.html) dataset from torchvision. It is still usefull ?

Fair point

For now, perhaps we only use ImageFolder for training and it will simplify things until we can share a public dataset

frgfm (changes requested):
In this case, there are a few things missing:
- removing openfire from datasets/__init__.py
- updating the reference script to use ImageFolder

MateoLostanlen:
Done :)

references/classification/README.md
```diff
@@ -19,19 +19,9 @@ pip install pyronear
...
-You can also use freely our open-source dataset
-
-```
-python train.py OpenFire/ --use-openfire --model rexnet1_0x --lr 1e-3 -b 16 --epochs 20 --opt radam --sched onecycle --device 0
-```
-
-### Wildfire
-
-This dataset is private, for now only Pyronear members can access it
+Train the model on your own dataset
```
frgfm, line 22: Perhaps we should specify the folder expected structure / hierarchy?

frgfm (approved):
Thanks a lot 🙂
Perhaps we should document the structure of the folder but it's nitpicking, no problem for me to move forward with this 👍

frgfm:
Overruled by #153 :) 
OpenFire is back :+1:

PR #129, opened 2021-10-25
fix: Fixed import for torch 1.10 release

MateoLostanlen:
Small PR to fix our [unitest](https://github.com/pyronear/pyro-vision/runs/4002060618?check_suite_focus=true)

The function [load_state_dict_from_url is now part of torch.hub](https://github.com/pytorch/pytorch/blob/f4dd88489a77ff9b300bf6f9b34c233ca82f76d7/torch/hub.py#L529)

We need to update the import accordingly

We must therefore set the minimal version of pytorch to 1.10


MateoLostanlen:
I wonder if setting the minimal version of pytorch to the last release is not too restrictive. @frgfm what do you think ?

It seems you have the same issue in holocron :)

frgfm:
> I wonder if setting the minimal version of pytorch to the last release is not too restrictive. @frgfm what do you think ?
> 
> It seems you have the same issue in holocron :)

Yup I agree, either I update it or do a conditional import by checking the library version :+1: 

Considering `0.10.0` is brand new, either we decide to:
- update it because there are features that we consider critical in this release
- add an upper constraint (<0.10.0)
- make the import conditional

I would go for the latter in this situation :+1:

MateoLostanlen:
It seems to be the best solution indeed, I opened [a PR](https://github.com/frgfm/Holocron/pull/146) on holocron to solve the problem.

frgfm (changes requested):
Actually the function is available in PyTorch since 1.1.0 (https://github.com/pytorch/pytorch/blob/v1.1.0/torch/hub.py#L385), and torchvision was only making a mirror for it up until the last release! So we only need:
```python
from torch.hub import load_state_dict_from_url
```

MateoLostanlen:
Hi @frgfm, thanks for catching this !

frgfm (dismissed):
Thanks for the edits!

frgfm:
About the unittest crash, I'll make a release of Holocron rather soon, we'll need to update the version constraint ;)

MateoLostanlen:
> 
> 
> About the unittest crash, I'll make a release of Holocron rather soon, we'll need to update the version constraint ;)

Ok Let me know when it's done I will wait to go further with this PR and set holocron minimal requirement (0.1.4 I guess).

frgfm:
> > About the unittest crash, I'll make a release of Holocron rather soon, we'll need to update the version constraint ;)
> 
> Ok Let me know when it's done I will wait to go further with this PR and set holocron minimal requirement (0.1.4 I guess).

To avoid waiting for the release, I suggest switching all pylocron requirements ([here](https://github.com/pyronear/pyro-vision/blob/master/requirements.txt#L7) and [here](https://github.com/pyronear/pyro-vision/blob/master/setup.py#L46)) from
```
pylocron>=0.1.3
```
to 
```
-e git+https://github.com/frgfm/Holocron.git#egg=pylocron
```

This will install the latest version from the git master branch instead of the release :+1:

MateoLostanlen:
Hi @frgfm, unfortunately we can' t use `e git+https://github.com/frgfm/Holocron.git#egg=pylocron ` in setup.py. A simple solution while waiting for the new holocron release is to install the requirements and then the package as I do [here](https://github.com/MateoLostanlen/pyro-vision/blob/a48daebad7d35b78123f4cf5cbdeaaf106512258/.github/workflows/builds.yml#L37).

The problem seems to be solved for the moment.

Only one test doesn't pass but it's always the same problem of openfire which has less and less image available. This problem will be solved after the merge of #121

.github/workflows/builds.yml
```diff
@@ -34,6 +34,7 @@ jobs:
       - name: Install package
         run: |
           python -m pip install --upgrade pip
+          pip install -r requirements.txt
```
frgfm: That line isn't needed: the next one will install all deps listed in `setup.py` along with the lib. And the latest pypi release of pylocron is a valid one, so I'd suggest to remove this

requirements.txt
```diff
@@ -4,4 +4,4 @@ torch>=1.8.0
 torchvision>=0.9.0
 tqdm>=4.20.0
 requests>=2.20.0
-pylocron>=0.1.3
+-e git+https://github.com/frgfm/Holocron.git#egg=pylocron
```
frgfm: would be safer to go for `pylocron>=0.2.0` and it would guarantee deps consistency between `requirements.txt` and `setup.py`

setup.py
```diff
@@ -43,7 +43,7 @@
     'torchvision>=0.9.0',
     'tqdm>=4.20.0',
     'requests>=2.20.0',
-    'pylocron>=0.1.3',
+    'pylocron',
```
frgfm: `pylocron>=0.2.0`

frgfm (commented):
Thanks! I added a few suggestions

MateoLostanlen:
Hi @frgfm, I think you did not understand. I propose here a temporary fix while you make your release because it blocks the pr on vision and engine for the moment. This modification allows to install holocron from the repo without waiting for the release. So you can take your time :)

MateoLostanlen:
ah ok you release 0.2.0, I will make the change then

MateoLostanlen:
all good :)

frgfm (approved):
All good!

PR #130, opened 2021-11-01
chore: Fixed documentation build and added CI jobs for doc sanity check

frgfm:
- one of sphinx deps `docutils` recently made a release that crashed new builds, so this adds a version constraint to avoid that latest release
- the documentation build was failing silently, so I added a CI job to ensure that at least the landing page was built properly in HTML

This has solved the issue on other projects of mine.

Any feedback is welcome!


MateoLostanlen (approved):
Thanks for the fix !

PR #131, opened 2021-11-28
Mlops on Test ardeche dataset

MateoLostanlen:
We have seen through our test phase in Ardeche the very poor generalization of the current model, especially on false positives. I created a [test dataset ](https://drive.google.com/file/d/13-5NJuVr29m0n3MG9wcYLjdgu2_2Hye0/view?usp=sharing)from a selection of images collected in Ardeche. As we can see here the performances of the current model on this dataset are very bad:
- Accuracy : 55 %
- Precision : 47 %
- Recall : 66 %

I propose to challenge our next model on this dataset. For that purpose I propose here a standardized evaluation of the model performances in a CI to follow at best the evolution of our performances.

In my opinion to obtain good performances on this dataset is now the main objective of the vision team


MateoLostanlen:
I will probably create an ardeche dataset on using openfirev2 template as soon as the [PR](https://github.com/pyronear/pyro-vision/pull/121) is merged

references/classification/test_ardeche.py
```diff
@@ -0,0 +1,54 @@
+from google_drive_downloader import GoogleDriveDownloader as gdd
+from pyrovision.models.rexnet import rexnet1_0x
+import torch
+from torchvision.datasets import ImageFolder
+from torch.utils.data import DataLoader
+from torchvision import transforms
+import json
+
+
```
frgfm, line 9: let's have a def main()

and an argparser :)

references/classification/test_ardeche.py
```diff
@@ -0,0 +1,54 @@
...
+# Download dataset
+gdd.download_file_from_google_drive(file_id='13-5NJuVr29m0n3MG9wcYLjdgu2_2Hye0',
+                                    dest_path='./ds.zip',
+                                    unzip=True)
+
```
frgfm, line 14: I'd suggest that the user passes the path to the data folder (can be downloaded from GDrive manually) to make this more flexible for future use.

frgfm (changes requested):
Thanks for the PR!

As discussed offline:
- I suggest updating a bit the evaluation script but keeping the spirit of it
- we can add some CI job to check that the script works (in another PR) but running it fully would be too long
- I'll make a few PRs to refactor the references folder to report results in a csv file

For this PR, I suggest to rename the evaluation script (not necessarily ardeche) to "evaluate.py" and make it a bit more flexible

frgfm:
With the recent refacto, I think we can close this PR

MateoLostanlen:
Yes we can !

PR #133, opened 2022-02-13
refactor: Refactored setup.py

frgfm:
- made sure the `.__version__` attribute is correctly set up dynamically
- added extra builds (meaning that `pip install -e .[docs]` will install the basic deps + those of the group I called "docs")

Any feedback is welcome!


MateoLostanlen (approved):
thanks for this fix !

PR #135, opened 2022-02-14
ci: Updated python version for builds

MateoLostanlen:
Python 3.6 is getting quite old (release in 2016) I think we can delete this test and add python 3.9 instead

[The build test](https://github.com/pyronear/pyro-vision/runs/5176415971?check_suite_focus=true) do not work anymore on mac


frgfm (approved):
Thanks! I was thinking to change this over the weekend :pray:

PR #136, opened 2022-02-14
remove dataset creation related modules

MateoLostanlen:
As discuss in #94 we need to reorganize the dataset folder

I have created a new [Pyro-dataset repo](https://github.com/pyronear/pyro-dataset). This new repo gather here all the necessary elements for dataset creation from video combining:
- Split the videos in sequences using our [Video Splitter](https://github.com/pyronear/VideoSplitter)
- Video sequence annotation from our [Video Annotator](https://github.com/pyronear/VideoAnnotator)
- Image dataset creation using this sequences and the annotation using [wildfire](https://github.com/pyronear/pyro-vision/blob/master/pyrovision/datasets/wildfire.py)
- Other usefull fonction in [video_utils](https://github.com/pyronear/pyro-vision/blob/master/pyrovision/datasets/video_utils.py) and [utils](https://github.com/pyronear/pyro-vision/blob/master/pyrovision/datasets/utils.py) modules

This way Pyro-vision become focus on model training

Moreover we remove the opencv dependencies witch can be annoying  

I need to waith the merge of #121 to remove [utils](https://github.com/pyronear/pyro-vision/blob/master/pyrovision/datasets/utils.py)


requirements.txt
```diff
@@ -1,7 +1,4 @@
-opencv-python>=3.4.5.20
-pandas>=0.25.2
 torch>=1.8.0
 torchvision>=0.9.0
-tqdm>=4.20.0
```
frgfm, line 5: we are still using it in datasets/utils I think

frgfm (commented):
Thanks! Only a small adjustment to make on the dependencies I think

requirements.txt
MateoLostanlen: I already move datasets/utils to pyro-dataset as it's dataset creation related

frgfm: Not sure we talk about the same thing, but as far as this branch is concerned, there is still an import of tqdm : https://github.com/MateoLostanlen/pyro-vision/blob/remove_dataset_modules/pyrovision/datasets/utils.py#L11

MateoLostanlen: Yes we are, but this module will be deleted with openfire, I can remove the dependency in the other PR

frgfm (approved):
Thanks for the edits :pray:

PR #137, opened 2022-03-13
ignore version file

MateoLostanlen:
Because version file if generated from setup, maybe it's cleaner to ignore it to avoid a useless untracked file


frgfm (approved):
Good idea!

PR #140, opened 2022-05-20
feat: Added codecarbon integration

frgfm:
- adds decorator integration of codecarbon
- updates requirements and gitignore

Any feedback is welcome!

cc @benoit-cty


MateoLostanlen:
Hi @frgfm thanks for this ! I do it work ? do we have an associated login ? or anyone cloning this repo will impact our score ?

frgfm:
Good questions, and that's why this hasn't been merged yet:
- one need a local config file, that hold the creds to upload this under "Data for Good - Pyronear" API access

The question now is what can we do and what do we want to do about people cloning the repo 🤔 

A suggestion I had in mind:
- public users : 2 options
    - we create a default cred that will be available publicly on the repo
    - we don't track it
- private/internal access : we use the config I told you about to upload all this under the same access

What do you think?

MateoLostanlen:
I think it's better not to track public users and to offer access to anyone who is close to the project and making them aware of gpu abuse.

benoit-cty:
Hello,
Your current code will work with no config file : the user emissions will be tracked but not send to the CodeCarbon API. Your user will have a local file `emissions.csv`. It's the default behavior of CodeCarbon.
So I think you are fine: When you are in an environment with proper config file, your emissions will go into the CodeCarbon API.

MateoLostanlen (dismissed):

frgfm:
In another PR, we can throw a warning when there is no config file (so that core team doesn't forget to set it when running a training)

PR #141, opened 2022-06-08
docs: Fixed URL in CONTRIBUTING

frgfm:
This PR fixes a broken URL in CONTRIBUTING.

Any feedback is welcome!


frgfm:
The CI job is doing weird things: the two other links mention do work. They're redirecting to something else and if I change it to the end URL, it still fails :man_shrugging: 

So I'll leave it as is

PR #163, opened 2022-09-16
Enhance conda meta.yaml

skupr-anaconda:
Actions:
1. Use `jinja2` template for the package name.
2. Add explicitly `pip` and `wheel` in `host`, see about wheel https://github.com/pyronear/pyro-vision/blob/6d88202fc80e8bd4c368018aacae9fa700091f43/pyproject.toml#L2
3. Improve a script
4. Remove whitespaces
5. Reorder `run` dependencies alphabetically
6. Add a commented `#- huggingface-hub>=0.4.0` to `run`, see https://github.com/pyronear/pyro-vision/blob/6d88202fc80e8bd4c368018aacae9fa700091f43/pyproject.toml#L38
7. Add `pip check` command
8. Fix license name, see https://spdx.org/licenses/
9. Add `license_family`

### DIFF
--- a/.conda/meta.yaml
+++ b/.conda/meta.yaml
@@ -4,30 +4,35 @@
 # {% set pyproject = load_file_data('../pyproject.toml', from_recipe_dir=True) %}
 # {% set project = pyproject.get('project') %}
 # {% set urls = pyproject.get('project', {}).get('urls') %}
+{% set name = "pyrovision" %}
+
 package:
-  name: pyrovision
+  name: {{ name|lower }}
   version: "{{ environ.get('BUILD_VERSION') }}"
 
 source:
-  fn: pyrovision-{{ environ.get('BUILD_VERSION') }}.tar.gz
-  url: ../dist/pyrovision-{{ environ.get('BUILD_VERSION') }}.tar.gz
+  fn: {{ name }}-{{ environ.get('BUILD_VERSION') }}.tar.gz
+  url: ../dist/{{ name }}-{{ environ.get('BUILD_VERSION') }}.tar.gz
 
 build:
+  number: 0
   noarch: python
-  script: python setup.py install --single-version-externally-managed --record=record.txt
+  script: {{ PYTHON }} setup.py install --single-version-externally-managed --record=record.txt
 
 requirements:
   host:
-    - python>=3.6, <4.0
+    - python>=3.6,<4.0
+    - pip
     - setuptools
-
+    - wheel
   run:
-    - python>=3.6, <4.0
-    - pytorch>=1.11.0, <2.0.0
-    - torchvision>=0.12.0, <1.0.0
+    - python>=3.6,<4.0
+    #- huggingface-hub>=0.4.0
+    - pylocron>=0.2.0,<1.0.0
+    - pytorch>=1.11.0,<2.0.0
+    - requests>=2.20.0,<3.0.0
+    - torchvision>=0.12.0,<1.0.0
     - tqdm>=4.1.0
-    - requests>=2.20.0, <3.0.0
-    - pylocron>=0.2.0, <1.0.0
 
 test:
   # Python imports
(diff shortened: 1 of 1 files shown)


.conda/meta.yaml
```diff
@@ -4,30 +4,35 @@
...
+    - pip
     - setuptools
-
+    - wheel
   run:
-    - python>=3.6, <4.0
-    - pytorch>=1.11.0, <2.0.0
-    - torchvision>=0.12.0, <1.0.0
+    - python>=3.6,<4.0
+    #- huggingface-hub>=0.4.0
```
frgfm: I understand it's in the pyproject.toml, but would you mind explaining why you remove it from here?
Is it because you default to pip install since there is no matching conda build?

frgfm (commented):
Thanks a lot @skupr-anaconda :pray: I added some questions in comments!

.conda/meta.yaml
skupr-anaconda: `huggingface-hub` is available only in the `Meepio` channel https://anaconda.org/Meepio/repo so `conda-build` can't install that package from the `main` channel.

frgfm: I see! But can't we simply add "Meepio" as a source channel for the build?

I'm trying to understand the reasoning for future decisions: why go for the pypi build if we can go for a conda build? (perhaps because meepio isn't the official HF conda channel?)

frgfm:
Oh and I just merged a PR that brought some conflict here. You'll need to merge master into your branch :/

.conda/meta.yaml
frgfm: if you don't mind, before moving on with the PR, would you mind sharing a bit more info on this @skupr-anaconda ? :)

I'm trying to better understand the reasoning for future similar situations

skupr-anaconda: It's not a usual situation for me too :)
If you can install your package without `huggingface-hub` you can simply add to the `test/commands` a new line:
` - pip install huggingface-hub`
In that way, you will be confident that all works correctly because `huggingface-hub` will be in the test environment.

frgfm: I don't get how it's linked, here is what I understood from your previous message:
- we can add a test command to see if hf hub can be installed
- however this doesn't mean that the conda build will be shipped with hf hub inside :thinking: 

How do you see things here? (Happy to merge but I really don't get the logic)

frgfm (commented):
Oops, I made a comment but didnt go through with the review!

.conda/meta.yaml
skupr-anaconda: `pyro-vision` will be shipped without `huggingface-hub`. 
I guess `pyro-vision` can be installed and run **with** and **without** `huggingface-hub`. 
You only test that `pyro-vision` is running correctly **with** it. 
After `conda install pyro-vision`  the end-user can install `huggingface-hub`, for example, by `pip install huggingface-hub`.

frgfm: Alright, I thought about only leaving the conda option if all deps can be installed that way, but I guess it helps already :+1: 
However, that does mean that someone installing pyrovision via conda will need to additionally install HF hub on their own (hence updating the installation instructions) if I understand this correctly?

skupr-anaconda: Yeah, you understand correctly

frgfm: Nice about my late answer @skupr-anaconda :upside_down_face: 
Final question: do you see a way to make sure that users, if they go for conda, don't have to go through additional installation steps? (because I see that as a regression for user installation otherwise)

skupr-anaconda: Yep, you can do it now because `huggingface_hub` (with **underscore**!) is already available on the default channel https://anaconda.org/main/huggingface_hub

frgfm: Oh that's perfect, would you mind updating your PR to add the dependency back please? :pray:  We can easily merge after this :)

skupr-anaconda: Done!

frgfm (approved):
Thank you very much Serhii :pray:

PR #164, opened 2022-09-18
fix: Fixed rexnet1_3x and rexnet1_5x and updated checkpoints

frgfm:
- fixes building of `rexnet1_3x` and `rexnet1_5x` (they were actually building a `rexnet1_0x` because of a typo)
- adds new features to training scipt: non-square image size, resize mode, updated augmentations, improved sample plotting, added gradient accumulation option, added an option to analyze training set image sizes and added a feature of Holocron to plot samples in the training set that yielded the highest loss
- retrained all rexnets & updated checkpoints
- updated the csv with performance
- fixed the resizing in API & demo

Any feedback is welcome!

### DIFF
--- a/api/app/vision.py
+++ b/api/app/vision.py
@@ -36,8 +36,8 @@ def preprocess_image(pil_img: Image.Image) -> np.ndarray:
         the resized and normalized image of shape (1, C, H, W)
     """
 
-    # Resizing
-    img = pil_img.resize(MODEL_CFG["input_shape"][-2:], Image.BILINEAR)
+    # Resizing (PIL takes (W, H) order for resizing)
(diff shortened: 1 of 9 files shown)


MateoLostanlen:
Hi @frgfm just a question did you use [readme parameters](https://github.com/pyronear/pyro-vision/tree/master/references/classification) for all models ? If not, it would be nice to add them to the performance table to keep track

frgfm:
@MateoLostanlen I kept everything, but the args are not always the same from one model to the other. So we'll have to think about how to keep track of that & share it

frgfm:
I just updated the readme instruction to display the command I use most often :+1:

PR #174, opened 2022-12-20
chore: Bumps pytest version to avoids CWE-1333

frgfm:
This PR bumps the pytest version to prevent a vulnerability spotted by the dependabot (https://github.com/pyronear/pyro-vision/security/dependabot/2).

Any feedback is welcome!


frgfm:
Same

